PR #334, opened 2015-05-21
numerous csw3 additions

tomkralidis:


tomkralidis:
FYI note that this PR breaks CSW 2.0.2 CITE compliance, which is rooted in an issue with the CSW 2.0.2 assertions.  This has been made known to OGC: https://github.com/opengeospatial/ets-csw202/issues/8

kalxas:
Since pycsw 1.x passes CITE we are ok.
For pycsw to continue passing the tests, OGC has to fix the test to take into account the CSW 3.0.0 spec

PR #351, opened 2015-06-28
adjust OpenSearch/Atom output

tomkralidis:
This PR brings the CITE CAT 3.0 testing to near completion:

- passed: 123
- failed: 1
- skipped: 0


kalxas:
The <atom:updated> tag is failing because that is hard-coded in the unit tests.
The atom unit tests need to ignore the actual value of this element.

tomkralidis:
Just sent another update to the tests.  Let's see what Travis says now.

Issue #149, opened 2013-04-29
Implement CSW 3.0


kalxas:
- [x] review specification
- [x] abstract codebase to deal with multiple versions
  - [x] get csw2 working first
- [x] implement request handling / new XML Schemas / validation
- [x] implement response handling
- [x] add tests to `tests/suites/csw3`
- [x] pass [OGC CITE CSW3 tests](https://github.com/opengeospatial/ets-cat30)
- [x] pass [OGC CITE CSW2 tests](https://github.com/opengeospatial/ets-csw202)
- [x] decide default version behaviour
- [x] update documentation
- [x] update http://demo.pycsw.org
- [x] update website

kalxas:
We have just branched master to https://github.com/geopython/pycsw/tree/csw3

tomkralidis:
old branch now at https://github.com/kalxas/pycsw/tree/csw3-old

tomkralidis:
Current status:

- passed: 115
- skipped: 1
- failed: 8

kalxas:
We are now passing all CITE tests for CSW3 :)

kalxas:
CSW3 support is merged in master now.

PR #354, opened 2015-06-30
Implement CSW 3.0 (#149)

tomkralidis:
Per #149


kalxas:
test merge here: https://github.com/kalxas/pycsw/commits/master

Issue #355, opened 2015-07-02
csw3.py conflicts with ebrim profile with default settings.


kalxas:
[Thu Jul 02 12:19:26 2015] [error] [client 176.58.229.210] Traceback (most recent call last):
[Thu Jul 02 12:19:26 2015] [error] [client 176.58.229.210]   File "/osgeo/demo.pycsw.org/pycsw/csw.wsgi", line 99, in application
[Thu Jul 02 12:19:26 2015] [error] [client 176.58.229.210]     status, contents = csw.dispatch_wsgi()
[Thu Jul 02 12:19:26 2015] [error] [client 176.58.229.210]   File "/osgeo/demo.pycsw.org/src/pycsw/pycsw/server.py", line 293, in dispatch_wsgi
[Thu Jul 02 12:19:26 2015] [error] [client 176.58.229.210]     return self.dispatch()
[Thu Jul 02 12:19:26 2015] [error] [client 176.58.229.210]   File "/osgeo/demo.pycsw.org/src/pycsw/pycsw/server.py", line 572, in dispatch
[Thu Jul 02 12:19:26 2015] [error] [client 176.58.229.210]     self.response = self.iface.getcapabilities()
[Thu Jul 02 12:19:26 2015] [error] [client 176.58.229.210]     return self.dispatch()
[Thu Jul 02 12:19:26 2015] [error] [client 176.58.229.210]   File "/osgeo/demo.pycsw.org/src/pycsw/pycsw/server.py", line 572, in dispatch
[Thu Jul 02 12:19:26 2015] [error] [client 176.58.229.210]     self.response = self.iface.getcapabilities()
[Thu Jul 02 12:19:26 2015] [error] [client 176.58.229.210]   File "/osgeo/demo.pycsw.org/src/pycsw/pycsw/ogc/csw/csw3.py", line 181, in getcapabilities
[Thu Jul 02 12:19:26 2015] [error] [client 176.58.229.210]     prof_val = self.parent.profiles['loaded'][prof].namespaces[prof_name]
[Thu Jul 02 12:19:26 2015] [error] [client 176.58.229.210] KeyError: 'ebrim'

tomkralidis:
For some reason ebrim setting causes error when in CSW 3. Works fine for CSW 2.

PR #356, opened 2015-07-02
Switching version to 2.0-dev

kalxas:


kalxas:
One test fails probably because there is a mismatch in versions between demo.pycsw.org and this branch.

tomkralidis:
@kalxas thanks! The travis-ci issue is unrelated and due to #355

PR #357, opened 2015-07-03
fix ebrim profile output for CSW3 (#355) and namespaces

tomkralidis:


kalxas:
I see 11 test errors in Travis.

PR #363, opened 2015-07-09
fixed #362 normalizing KVP in dispatch method

### LINKED ISSUE #362: server.dispatch() does not normalize KVP parameter names
The `pycsw.server.Csw.dispatch` method does not normalize the names of KVP parameters. Instead it relies on the `dispatch_cgi` (line 229) and `dispatch_wsgi` (283) methods to do this. It would be cleaner to do this inside the `dispatch` method which is the next step in the request processing pipeline. 

Advantages:

* DRYier - We eliminate duplicate code
* It is easier to extend pycsw by calling server.dispatch

An example:

My current use case is integrating PyCSW with django. Inside my django view I instantiate the csw server and initialize it with the appropriate settings. Now I want to use the `server.dispatch` method, but I must remember to pre_process KVP param names to lower case.

This seems like an easy fix and I can supply a Pull Request with the modification in the near future

ricardogsilva:
Hi all, I'd like to contribute the following bugfix for issue #362 to pycsw. I confirm that my contributions to pycsw will be compatible with the pycsw license guidelines at the time of contribution.

This pull request fixes issue #362 by performing normalization of a GET request's KVP parameters in the `pycsw.server.dispatch` method.

### DIFF
--- a/pycsw/server.py
+++ b/pycsw/server.py
@@ -226,7 +226,7 @@ class Csw(object):
             (self.environ['HTTP_HOST'], self.environ['REQUEST_URI'])
             LOGGER.debug('Request type: GET.  Request:\n%s\n', self.request)
             for key in cgifs.keys():
-                self.kvp[key.lower()] = cgifs[key].value
+                self.kvp[key] = cgifs[key].value
 
         return self.dispatch()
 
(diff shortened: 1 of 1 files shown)


tomkralidis:
Thanks @ricardogsilva !

PR #368, opened 2015-08-14
refactored the cgi wrapper to use code from wsgi

ricardogsilva:
This pull request refactors the `csw.py` cgi wrapper in order to reuse the code from the WSGI wrapper.

Since pycsw already has a WSGI wrapper that makes use of python's `wsgiref` library and since wsgiref includes a handler for CGI, it is possible to reuse the same wrapper code to handle both WSGI and CGI. This makes it possible to be a bit DRYer ;)

Changed stuff:

* Removed most code from the CGI wrapper script. yay!

* Removed the `pycsw.server.dispatch_cgi` method. We do not need it now that CGI is being handled by WSGI

* The `csw.wsgi` module had to be moved inside the `pycsw` package in order to make it be importable in a robust way (we do not need to mess with sys.path if pycsw is already installed and this module is a part of it)

* While moving `csw.wsgi` I've also renamed it to `wsgi.py` in order to make it a bit more standard

* Updated the installation docs

* Updated the opensuse build recipe, but I must admit that I'm not sure of the changes, since this is not something I had done before

* Updated the `pavement.py` file in order to update the new path of the wsgi wrapper

### DIFF
--- a/csw.py
+++ b/csw.py
@@ -29,56 +29,14 @@
 #
 # =================================================================
 
-# CGI wrapper for pycsw
+"""
+A CGI wrapper for pycsw that reuses code from the wsgi wrapper.
+"""
 
-import os
-import sys
-from StringIO import StringIO
-from pycsw import server
+from wsgiref.handlers import CGIHandler
 
-CONFIG = 'default.cfg'
-GZIP = False
+from pycsw.wsgi import application
 
-if 'PYCSW_CONFIG' in os.environ:
-    CONFIG = os.environ['PYCSW_CONFIG']
-if os.environ['QUERY_STRING'].lower().find('config') != -1:
-    for kvp in os.environ['QUERY_STRING'].split('&'):
-        if kvp.lower().find('config') != -1:
-            CONFIG = kvp.split('=')[1]
 
-if ('HTTP_ACCEPT_ENCODING' in os.environ and
-        os.environ['HTTP_ACCEPT_ENCODING'].find('gzip') != -1):
-    # set for gzip compressed response
-    GZIP = True
-
-# get runtime configuration
-CSW = server.Csw(CONFIG)
-
-# set compression level
-if CSW.config.has_option('server', 'gzip_compresslevel'):
-    GZIP_COMPRESSLEVEL = \
-        int(CSW.config.get('server', 'gzip_compresslevel'))
-else:
-    GZIP_COMPRESSLEVEL = 0
-
-# go!
-STATUS, OUTP = CSW.dispatch_cgi()
-
-sys.stdout.write("Content-Type:%s\r\n" % CSW.contenttype)
-
-if GZIP and GZIP_COMPRESSLEVEL > 0:
-    import gzip
-
-    BUF = StringIO()
-    GZIPFILE = gzip.GzipFile(mode='wb', fileobj=BUF,
-                             compresslevel=GZIP_COMPRESSLEVEL)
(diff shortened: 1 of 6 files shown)


tomkralidis:
@ricardogsilva thanks for this.  Very nice work here to really simplify things.  Comments:

1. great to see the CGI wrapper simplified.  I assume the `GZIP_COMPRESSLEVEL` stuff handled in the same way?
1. I will defer opensuse issues to @kalxas 
1. I'm assuming this has been tested (we don't have CGI-based tests, perhaps something we need to add to our unit tests in the future [outside the scope of this PR])
1. moving `csw.wsgi` to the library: while I'm not opposed to this per se, I'm not sure of the implications of this change w.r.t. infrastructure and packaging.  I believe currently we have `python-pycsw` as the library package and `python-pycsw-wsgi` and `python-pycsw-cgi` for the wrappers but @kalxas has a better understanding of our packaging/build story so I defer to @kalxas here

Once we clarify/verify all is good with comment 4, I'm +1 for this awesome contribution.

ricardogsilva:
Hi @tomkralidis 

Regarding your comments:

1. I did not touch the wsgi script (other than renaming and moving it) so I assume gzip stuff is handled just as before. I've been testing this PR using python's built in cgi server. This server does not seem to process the `Accept-Encoding: gzip, deflate` header, so I have not tested it. Since the code that checks the gzip stuff in the old version of `csw.py` is the same as the one on the `wsgi.py` module, I assume things to work.

  The way in which I have been testing is:

  * Somewhere under your home directory create a `cgi-bin` directory and place a symlink to `csw-py` inside it. Navigate to the directory above `cgi-bin` and start python's CGI server:

  ```bash
  cd ~
  mkdir cgi-bin
  ln -s /home/ricardo/dev/pycsw/pycsw/csw.py cgi-bin/csw.py
  python -m CGIHTTPServer
  ```

  * Now make some CSW requests to `http:localhost:8000/cgi-bin/csw.py`

2. Yes, the opensuse part of the PR is just me hoping that it works. It needs review from someone experienced in packaging

3. As reported in 1. I've been testing this by performing some requests using python's builtin CGI server. I tried running the test suites, but realized that CGI is not being tested with them. As such I've hacked the `pavement.py` file to use my testing CGI server instead. The tests ran and I got 94.37% success (13 tests fail). This is the same result as when I am running the tests with the master branch, so everything seems to work OK.

  I agree that adding cgi to the automated tests would be nice. however, it make the testing (even) longer... I can tidy up my hack into another PR, if it seems appropriate

4. Allright, lets wait for feedback from @kalxas ;)

tomkralidis:
@ricardogsilva comments:

1. `GZIP_COMPRESSLEVEL`: you are right, this flows through
1. defer to @kalxas 
1. nice, let's leave CGI support in the unit tests for another PR/issue.  Feel free to open an [issue](https://github.com/geopython/pycsw/issues/new) on this one if you want
1. defer to @kalxas

tomkralidis:
@ricardogsilva P.S. I am getting the same behaviour as you, 13 tests fail consistently in my env, but not on travis-ci.

kalxas:
Hi all,

Regarding packaging, no issues here, we will review packaging for pycsw 2.0 (both rpm and deb).

kalxas:
P.S. I will run some tests here too and merge this PR

kalxas:
All tests passed here, merging.
Thanks @ricardogsilva

PR #376, opened 2015-09-11
Python 3 support.

ocefpaf:
This is not ready to be merged.  This PR implements:

- Preliminary Python3 support
- Travis-CI container (faster) builds

@QuLogic can you take a look at this?  You are one of the best Python2/3 conversion experts I known :wink: 

Note that we ca use `six.iteritems` instead simply changing `iteritems()` to `items`.  This is just a first draft to get things going.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -5,6 +5,7 @@ sudo: false
 python:
   - "2.6"
   - "2.7"
+  - "3.4"
 
 addons:
   apt:
@@ -28,5 +29,4 @@ notifications:
   irc:
     channels:
       - "irc.freenode.org#pycsw"
-      #- "irc.freenode.org#geopython"
     use_notice: true
--- a/pycsw/core/metadata.py
+++ b/pycsw/core/metadata.py
@@ -28,9 +28,11 @@
 #
 # =================================================================
 
+from __future__ import (absolute_import, division, print_function)
+
 import logging
 import uuid
-from urlparse import urlparse
+from six.moves.urllib.parse import urlparse
 from owslib.util import build_get_url
 from geolinks import sniff_link
 from pycsw.core import util
@@ -44,7 +46,7 @@ def parse_record(context, record, repos=None,
     ''' parse metadata '''
 
     if identifier is None:
-        identifier = uuid.uuid4().get_urn()
+        identifier = uuid.uuid4().urn
 
     # parse web services
     if (mtype == 'http://www.opengis.net/cat/csw/2.0.2' and
@@ -73,7 +75,7 @@ def parse_record(context, record, repos=None,
     elif mtype == 'http://www.opengis.net/wms':  # WMS
         LOGGER.debug('WMS detected, fetching via OWSLib')
         return _parse_wms(context, repos, record, identifier)
-     
+
     elif mtype == 'http://www.opengis.net/wps/1.0.0':  # WPS
         LOGGER.debug('WPS detected, fetching via OWSLib')
         return [_parse_wps(context, repos, record, identifier)]
@@ -219,7 +221,7 @@ def _parse_csw(context, repos, record, identifier, pagesize=10):
                            maxrecords=pagesize, outputschema=csw_outputschema, esn='full')
         except Exception as err:  # this is a CSW, but server rejects query
             raise RuntimeError(md.response)
-        for k, v in md.records.iteritems():
+        for k, v in md.records.items():
             # try to parse metadata
             try:
                 LOGGER.debug('Parsing metadata record: %s', v.xml)
@@ -247,7 +249,7 @@ def _parse_waf(context, repos, record, identifier):
         tree = etree.fromstring(content, parser=parser)
     except Exception as err:
         raise Exception('Could not parse WAF: %s' % str(err))
-        
+
     up = urlparse(record)
     links = []
 
@@ -333,8 +335,8 @@ def _parse_wms(context, repos, record, identifier):
     _set(context, serviceobj, 'pycsw:Links', '^'.join(links))
     _set(context, serviceobj, 'pycsw:XML', caps2iso(serviceobj, md, context))
 
-    recobjs.append(serviceobj) 
-         
+    recobjs.append(serviceobj)
+
     # generate record foreach layer
 
     LOGGER.debug('Harvesting %d WMS layers' % len(md.contents))
@@ -765,7 +767,7 @@ def _parse_fgdc(context, repos, exml):
     if md.idinfo.datasetid is not None:  # we need an identifier
         _set(context, recobj, 'pycsw:Identifier', md.idinfo.datasetid)
     else:  # generate one ourselves
-        _set(context, recobj, 'pycsw:Identifier', uuid.uuid1().get_urn())
+        _set(context, recobj, 'pycsw:Identifier', uuid.uuid1().urn)
 
     _set(context, recobj, 'pycsw:Typename', 'fgdc:metadata')
     _set(context, recobj, 'pycsw:Schema', context.namespaces['fgdc'])
@@ -879,7 +881,7 @@ def _parse_iso(context, repos, exml):
 
         if len(md.identification.resourcelanguage) > 0:
             _set(context, recobj, 'pycsw:ResourceLanguage', md.identification.resourcelanguage[0])
- 
+
         if hasattr(md.identification, 'bbox'):
             bbox = md.identification.bbox
         else:
@@ -904,16 +906,16 @@ def _parse_iso(context, repos, exml):
             all_orgs = set([item.organization for item in md.identification.contributor if hasattr(item, 'organization') and item.organization is not None])
             _set(context, recobj, 'pycsw:Contributor', ';'.join(all_orgs))
 
-        if (hasattr(md.identification, 'contact') and 
+        if (hasattr(md.identification, 'contact') and
             len(md.identification.contact) > 0):
             all_orgs = set([item.organization for item in md.identification.contact if hasattr(item, 'organization') and item.organization is not None])
             _set(context, recobj, 'pycsw:OrganizationName', ';'.join(all_orgs))
 
         if len(md.identification.securityconstraints) > 0:
-            _set(context, recobj, 'pycsw:SecurityConstraints', 
+            _set(context, recobj, 'pycsw:SecurityConstraints',
             md.identification.securityconstraints[0])
         if len(md.identification.accessconstraints) > 0:
-            _set(context, recobj, 'pycsw:AccessConstraints', 
+            _set(context, recobj, 'pycsw:AccessConstraints',
             md.identification.accessconstraints[0])
         if len(md.identification.otherconstraints) > 0:
             _set(context, recobj, 'pycsw:OtherConstraints', md.identification.otherconstraints[0])
@@ -951,7 +953,7 @@ def _parse_iso(context, repos, exml):
         _set(context, recobj, 'pycsw:ServiceTypeVersion', md.serviceidentification.version)
 
         _set(context, recobj, 'pycsw:CouplingType', md.serviceidentification.couplingtype)
-   
+
     service_types = []
     for smd in md.identificationinfo:
         if smd.identtype == 'service' and smd.type is not None:
@@ -959,14 +961,14 @@ def _parse_iso(context, repos, exml):
 
     _set(context, recobj, 'pycsw:ServiceType', ','.join(service_types))
 
-        #if len(md.serviceidentification.operateson) > 0: 
-        #    _set(context, recobj, 'pycsw:operateson = VARCHAR(32), 
-        #_set(context, recobj, 'pycsw:operation VARCHAR(32), 
-        #_set(context, recobj, 'pycsw:operatesonidentifier VARCHAR(32), 
-        #_set(context, recobj, 'pycsw:operatesoname VARCHAR(32), 
+        #if len(md.serviceidentification.operateson) > 0:
+        #    _set(context, recobj, 'pycsw:operateson = VARCHAR(32),
+        #_set(context, recobj, 'pycsw:operation VARCHAR(32),
+        #_set(context, recobj, 'pycsw:operatesonidentifier VARCHAR(32),
+        #_set(context, recobj, 'pycsw:operatesoname VARCHAR(32),
 
 
-    if hasattr(md.identification, 'dataquality'):     
+    if hasattr(md.identification, 'dataquality'):
         _set(context, recobj, 'pycsw:Degree', md.dataquality.conformancedegree)
         _set(context, recobj, 'pycsw:Lineage', md.dataquality.lineage)
         _set(context, recobj, 'pycsw:SpecificationTitle', md.dataquality.specificationtitle)
@@ -1102,6 +1104,6 @@ def caps2iso(record, caps, context):
     apiso_obj = APISO(context.model, context.namespaces, context)
     apiso_obj.ogc_schemas_base = 'http://schemas.opengis.net'
     apiso_obj.url = context.url
-    queryables = dict(apiso_obj.repository['queryables']['SupportedISOQueryables'].items() + apiso_obj.repository['queryables']['SupportedISOQueryables'].items())
+    queryables = dict(apiso_obj.repository['queryables']['SupportedISOQueryables'].items())
     iso_xml = apiso_obj.write_record(record, 'full', 'http://www.isotc211.org/2005/gmd', queryables, caps)
     return etree.tostring(iso_xml)
--- a/pycsw/core/util.py
+++ b/pycsw/core/util.py
@@ -30,14 +30,17 @@
 #
 # =================================================================
 
+from __future__ import (absolute_import, division, print_function)
+
 import time
 import datetime
 import logging
-import urllib2
+from six.moves.urllib.request import Request, urlopen
 from shapely.wkt import loads
 from owslib.util import http_post
 from pycsw.core.etree import etree
 
+from six import text_type
 LOGGER = logging.getLogger(__name__)
 
 #Global variables for spatial ranking algorithm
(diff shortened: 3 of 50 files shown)


tomkralidis:
The less new dependencies the better :)

QuLogic:
Is there a reason you added `print()` but no `__future__` statements?

pycsw/core/metadata.py
```diff
@@ -1102,6 +1105,6 @@ def caps2iso(record, caps, context):
     apiso_obj = APISO(context.model, context.namespaces, context)
     apiso_obj.ogc_schemas_base = 'http://schemas.opengis.net'
     apiso_obj.url = context.url
-    queryables = dict(apiso_obj.repository['queryables']['SupportedISOQueryables'].items() + apiso_obj.repository['queryables']['SupportedISOQueryables'].items())
+    queryables = dict(list(apiso_obj.repository['queryables']['SupportedISOQueryables'].items()) * 2)
```
QuLogic: What does this do? The key is unique, so what effect does the second copy have?

pycsw/core/util.py
```diff
@@ -33,11 +33,15 @@
...
-import urllib2
+try:
+    from urllib2 import Request, urlopen
+except:
+    from urllib.request import Request, urlopen
 from shapely.wkt import loads
 from owslib.util import http_post
 from pycsw.core.etree import etree
 
+from six import text_type
```
QuLogic: You still have a `six` dependency here (not that I care either way.)

pycsw/core/util.py
```diff
@@ -284,7 +288,7 @@ def transform_mappings(queryables, typename, reverse=False):
     if reverse:  # from csw:Record
         for qbl in queryables.keys():
             if qbl in typename.values():
-                tmp = [k for k, v in typename.iteritems() if v == qbl][0]
+                tmp = [k for k, v in typename.items() if v == qbl][0]
```
QuLogic: You could optimize the loop by using `next(k for k, v in typename.items() if v == qbl)`, _but_ keep in mind that dictionary iteration does not guarantee sort order unless this one is guaranteed to be an `OrderedDict`.

pycsw/core/util.py
```diff
@@ -302,9 +306,9 @@ def get_anytext(bag):
     """
 
     if isinstance(bag, list):  # list of words
-        return ' '.join(filter(None, bag)).strip()
+        return ' '.join([_f for _f in bag if _f]).strip()
     else:  # xml
-        if isinstance(bag, unicode) or isinstance(bag, str):  # not serialized yet
+        if isinstance(bag, text_type) or isinstance(bag, str):  # not serialized yet
```
QuLogic: This is basically `isinstance(bag, six.string_types)`.

pycsw/oaipmh.py
```diff
@@ -183,7 +183,7 @@ def response(self, response, kvp, repository, server_url):
             etree.SubElement(node, util.nspath_eval('oai:error', self.namespaces), code='badArgument').text = 'Missing \'verb\' parameter'
             return node
 
-        if kvp['verb'] not in self.request_model.keys():
+        if kvp['verb'] not in list(self.request_model.keys()):
```
QuLogic: The `list` is not necessary here, nor is the `.keys()`.

pycsw/oaipmh.py
```diff
@@ -197,11 +197,11 @@ def response(self, response, kvp, repository, server_url):
             if 'metadataprefix' not in kvp:
                 etree.SubElement(node, util.nspath_eval('oai:error', self.namespaces), code='badArgument').text = 'Missing metadataPrefix parameter'
                 return node
-            elif kvp['metadataprefix'] not in self.metadata_formats.keys():
+            elif kvp['metadataprefix'] not in list(self.metadata_formats.keys()):
```
QuLogic: The `list` is not necessary here, nor is the `.keys()`.

pycsw/ogc/csw/csw2.py
```diff
@@ -33,10 +33,15 @@
...
 import sys
 import cgi
-from urllib2 import quote, unquote
-import urlparse
-from cStringIO import StringIO
-from ConfigParser import SafeConfigParser
+try:
+    from urllib2 import quote, unquote
+except ImportError:
+    from urllib.parse import quote, unquote
```
QuLogic: Since there's a `six` import just below here, you can just import this from `six.moves` without the `try/except`.

pycsw/ogc/csw/csw2.py
```diff
@@ -33,10 +33,15 @@
...
+from six import StringIO
+try:
+    from ConfigParser import SafeConfigParser
+except ImportError:
+    from configparser import SafeConfigParser
```
QuLogic: This one can also be pulled from `six`.

pycsw/ogc/csw/csw3.py
```diff
@@ -514,9 +519,8 @@ def getdomain(self):
                     operation, parameter = pname.split('.')
                 except:
                     return node
-                if (operation in self.parent.context.model['operations'].keys() and
-                    parameter in
-                    self.parent.context.model['operations'][operation]['parameters'].keys()):
+                if (operation in list(self.parent.context.model['operations'].keys()) and
+                    parameter in list(self.parent.context.model['operations'][operation]['parameters'].keys())):
```
QuLogic: Can drop the `list` and `.keys()`, and this second line is not indented enough for PEP8.

pycsw/ogc/csw/csw3.py
```diff
@@ -707,7 +711,7 @@ def getrecords(self):
         # check elementname's
         if 'elementname' in self.parent.kvp:
             for ename in self.parent.kvp['elementname']:
-                enamelist = self.parent.repository.queryables['_all'].keys()
+                enamelist = list(self.parent.repository.queryables['_all'].keys())
```
QuLogic: This variable is not even necessary since both `list` and `keys()` are not.

pycsw/ogc/csw/csw3.py
```diff
@@ -942,7 +946,7 @@ def getrecords(self):
 
                         searchresults.append(self._write_record(
                         res, self.parent.repository.queryables['_all']))
-                    elif self.parent.kvp['outputschema'] in self.parent.outputschemas.keys():  # use outputschema serializer
+                    elif self.parent.kvp['outputschema'] in list(self.parent.outputschemas.keys()):  # use outputschema serializer
```
QuLogic: Can drop `list` and `.keys()`.

pycsw/ogc/csw/csw3.py
```diff
@@ -1121,7 +1125,7 @@ def getrecordbyid(self, raw=False):
                     ['mappings']['csw:Record'], reverse=True)
 
                 node = self._write_record( result, self.parent.repository.queryables['_all'])
-            elif self.parent.kvp['outputschema'] in self.parent.outputschemas.keys():  # use outputschema serializer
+            elif self.parent.kvp['outputschema'] in list(self.parent.outputschemas.keys()):  # use outputschema serializer
```
QuLogic: Can drop `list` and `.keys()`.

pycsw/ogc/csw/csw3.py
```diff
@@ -2049,9 +2053,9 @@ def resolve_nsmap(self, list_):
             LOGGER.debug(tn)
             if tn.find(':') != -1:  # resolve prefix
                 prefix = tn.split(':')[0]
-                if prefix in nsmap.keys():  # get uri
+                if prefix in list(nsmap.keys()):  # get uri
```
QuLogic: Can drop `list` and `.keys()`.

pycsw/ogc/csw/csw3.py
```diff
@@ -2049,9 +2053,9 @@ def resolve_nsmap(self, list_):
...
                     uri = nsmap[prefix]
-                    newprefix = self.parent.context.namespaces.keys()[self.parent.context.namespaces.values().index(uri)]
+                    newprefix = list(self.parent.context.namespaces.keys())[list(self.parent.context.namespaces.values()).index(uri)]
```
QuLogic: Since there's no order to dictionary iteration, I don't think you should rely on `keys()` and `values()` returning lists in the same order. This should probably be re-written like the one a little bit above I commented on earlier.

pycsw/server.py
```diff
@@ -511,7 +520,7 @@ def dispatch(self, writer=sys.stdout, write_headers=True):
 
                 # test request
                 if self.kvp['request'] not in \
-                    self.context.model['operations'].keys():
+                    list(self.context.model['operations'].keys()):
```
QuLogic: Can drop `list` and `.keys()`.

.travis.yml
```diff
@@ -1,21 +1,23 @@
...
 
 install:
-  - sudo apt-get install -y libgeos-c1
   - pip install . --use-mirrors
-  - pip install -r ./requirements-standalone.txt
-  - pip install -r ./requirements-dev.txt
-  - python ./setup.py -q install
-
-before_script:
-  - sudo add-apt-repository ppa:gcpp-kalxas/ppa-tzotsos -y
```
kalxas: These lines here should not be removed, the tests will fail with old libxml...

requirements.txt
```diff
@@ -1,5 +1,6 @@
 geolinks==0.2.0
 lxml==3.3.5
-OWSLib<0.9.0
```
kalxas: Tests did not pass with OWSLib 0.9.x in the past.
Has something changed with that?

.travis.yml
kalxas: We switched to new libxml to be able to validate requests here:
https://github.com/geopython/pycsw/commit/7ab7a4051afa7acf0875fc51852a6141715aed56
https://github.com/geopython/pycsw/commit/32b1dc05f4d219ba7f7a20c125b050c0056e2dc9

Has something changed with Travis? Does it use Ubuntu Trusty now?

pycsw/core/util.py
ocefpaf: You already added six as a dependency for OWSLib, which is a dependency for pycws,
so I decided to go the six way...

pycsw/core/util.py
ocefpaf: I am unfamiliar with the code to know if an un-ordered dict is an issue here.  I will try and let's hope the tests a comprehensive enough.

pycsw/core/metadata.py
ocefpaf: We need someone with a deeper knowledge of `pycsw` to answer that...

pycsw/core/util.py
QuLogic: Sure, but @tomkralidis was mentioning less dependencies...

tomkralidis: Yes the less dependencies the better. Even if OWSLib brings those demos in at least pycsw will not have to keep account of them.

pycsw/core/util.py
tomkralidis: If the unit test pass then this is ok

.travis.yml
QuLogic: Tests (on Python 2) seemed okay without it. Travis does not yet use Trusty unless you request it.

pycsw/core/util.py
QuLogic: The trouble with non-guaranteed order is sometimes it will work and sometimes it won't.

.travis.yml
ocefpaf: They where OK.  I just added some extra commits that broke it.  `libxml` and OWSLib are not an issue.

kalxas: There has to be a change in libxml used by Travis, since the libxml version in Precise fails to parse GML. This is why I used my own backport here.

pycsw/core/util.py
ocefpaf: No problem.  We can workaround them or use a `six.py` copy with the parts we need.  (Six was actually intended to be used that way.)  Let me just work some `encode/decode` details out and them I figure out the best strategy.

pycsw/core/metadata.py
tomkralidis: Looks like a vintage copy/paste typo :)

Having said this, if we take it out, the tests will tell us whether this matters or not.

.travis.yml
tomkralidis: It looks like moving to Travis' [container-based infrastructure](http://docs.travis-ci.com/user/workers/container-based-infrastructure/) results in the libxml2 upgrade, i.e.:

- master: https://travis-ci.org/geopython/pycsw/jobs/78919311
- this PR: https://travis-ci.org/geopython/pycsw/jobs/80046171#L83

Having said this I'm +1 with pushing the `.travis.yml` changes immediately in advance of this PR (or even as an isolated PR), given it removes the requirement to manage our own libxml2 for the tests to correctly pass.

Thoughts?

pycsw/core/util.py
tomkralidis: Forgive my Python 3 ignorance here, but if pycsw doesn't import six, and OWSLib does, can pycsw still use it?  If pycsw doesn't use six, what are the implications?

If things are too painful without using six, and six has packaging distribution support (which @kalxas agrees to), then I'm not against using it per se.  Is using six a temporary solution?

Again, apologies for my Python 3 ignorance, and huge thanks to @ocefpaf and @QuLogic to help bring Python 3 into pycsw.

.travis.yml
kalxas: That explains the change in behaviour. I am also guessing that libxml 2.8 included in this new environment has parts of 2.9 backported...

ocefpaf: >  I'm +1 with pushing the `.travis.yml` changes immediately in advance of this PR.

Makes sense.  I actually should not have mixed things here.  (Also, this PR might take a little more effort than I thought to make better use of `six.moves` and Ordereddicts.)

Done in #377

pycsw/core/util.py
ocefpaf: > pycsw doesn't import six, and OWSLib does, can pycsw still use it?

Yes, but it would be better to be explicit and import it.  The alternative, to avoid adding an extra dependency, is to drop the `six.py` file there.  That **is** one of the recommended ways to use six.


> If pycsw doesn't use six, what are the implications?

It is just a little bit harder to  make code compliant with both python 2 and 3.  Six also makes the code more readable.

kalxas: I am ok with adding six as a dependency. Copying the six.py file in our code base would need some licensing review etc.

tomkralidis:
Suggest we include in requirements.txt, not inline.

.travis.yml
tomkralidis: Are any of the tests failing which causes OrderedDict to be required?

tests/run_tests.py
```diff
@@ -50,6 +52,7 @@ def plural(num):
 
 def get_validity(sexpected, sresult, soutfile, force_id_mask=False):
     """Decipher whether the output passes, fails, or initializes"""
+    sresult = sresult
```
tomkralidis: is this required?

tomkralidis:
Looks like the Python 3 tests fail in the test runner: https://travis-ci.org/geopython/pycsw/jobs/80099911#L1132

tests/run_tests.py
ocefpaf: Nope.  That is a left-over from my `decode`/`encode` whack-a-mole.  I need to study the code base a little bit more to efficiently encode things at the beginning and decode at the end to avoid messing up things.

pycsw/core/util.py
ocefpaf: > I am ok with adding six as a dependency.

That is better.  If one day `OWSLib` drops `six` than `pycsw` would be broken.  (Explicit is better than implicit.)

> Copying the six.py file in our code base would need some licensing review etc.

I believe that would be OK, but `six` is a lightweight dependency.

pycsw/core/util.py
ocefpaf: Done!  It is working just fine.  I guess only one match is expected.  (Thanks @QuLogic for the attention to detail!)

pycsw/core/util.py
tomkralidis: I'm +1 for adding `six` as a dep in `requirements.txt`.

pycsw/core/metadata.py
ocefpaf: Done.  This now passes on py26/27.

pycsw/core/util.py
ocefpaf: Using six everywhere but `pavement.py`.  I do not know how Paver works, but six did not get imported from there.

ocefpaf:
> Suggest we include in requirements.txt, not inline.

Not sure what you mean.  I kind of got lost in all the comments.

> Looks like the Python 3 tests fail in the test runner: https://travis-ci.org/geopython/pycsw/jobs/80099911#L1132

I fixed that.  Now all the tests fail on Python 3 :stuck_out_tongue_winking_eye: 

There are a few encode/decode to do. Also I guess some dicts and/or json dumps must be ordered.  I see several test failure due to out of xml.

We can continue on this PR (1) or merge this, since I manage to not break python 2, and open a new one to address the remaining issues (2).

Note that I would like to do (1), but I do not know when I will be able to get back to this and (2) might be appealing if someone else is willing to pick up from where I stopped.

kalxas:
We could make a Python_3 branch and move this work there, but it cannot be merged into master IMHO

tomkralidis:
cc @ricardogsilva 

@ocefpaf thanks.  We are close.  Looks like issues are bound to the following:

- varying order of XML root element namespaces declarations
- as noted, varying order of dicts and json 

AFAICT looks like most, if not all, errors can be solved by enforcing order in the codebase.  Making these changes will also require close inspection and update of the expected results which will naturally change.

I agree with @kalxas in making a `py3k` branch in pycsw proper.

ocefpaf:
> AFAICT looks like most, if not all, errors can be solved by enforcing order in the codebase. Making these changes will also require close inspection and update of the expected results which will naturally change.

Yep.  That is why I am stepping back.  The day job is calling and this will take more time than I thought.  Sorry.

> I agree with @kalxas in making a `py3k` branch in pycsw proper.

OK.  I think I can point this PR to a py3k branch `upstream`, but I am not sure if I can create that branch.

tomkralidis:
@ocefpaf branch created for PR: https://github.com/geopython/pycsw/tree/py3k

ocefpaf:
Done in #379.  Closing this one.

Issue #213, opened 2014-01-14
remove unused namespaces from response XML


tomkralidis:
Given all the namespaces we init, the resulting XML output ends being mostly/all namespace declarations in the root element.  Some of which are not required given the context of the request/response.

lxml's `etree.cleanup_namespaces` (http://lxml.de/api/lxml.etree-module.html#cleanup_namespaces) is a convenience function which discards unused namespaces and results in leaner, smarter XML output.

tomkralidis:
Some benchmarks (invoking a `GetCapabilities` request)

|  | size (KB) | average response time based on 10 requests (ms)  |
---- | ------- | ------------ | -- | ---
without `etree.cleanup_namespaces` | 15.37 | 170.4
with `etree.cleanup_namespaces` | 14.67 | 191.2
difference | -4.55%| +10.88%

kalxas:
Thanks for the benchmark @tomkralidis 
I think this slowdown is significant.

kalxas:
Merged in csw3 branch (which will lead to master)
I will keep this ticket open to add a switch in configuration (turning on/off this feature)

kalxas:
This commit has caused problems to CITE testing and has been reverted.

tomkralidis:
There is an issue when using `etree.cleanup_namespaces` against QName type element content:

https://mailman-mail5.webfaction.com/pipermail/lxml/2015-May/007512.html

We need to leave this as is until we find a workaround given CITE testing.

kalxas:
https://github.com/lxml/lxml/pull/167

tomkralidis:
lxml 3.5.0 has been released which [includes](https://github.com/lxml/lxml/blob/master/CHANGES.txt#L25) functionality pycsw can leverage to cover this issue.

@kalxas any objections/concerns to bumping lxml to 3.5.0 in master?

kalxas:
Requiring such new version of lxml might solve this issue, but we should expect the packaging side to have problems: most distributions will take time/years to reach this lxml version.

kalxas:
Anyway, since we are already strict on the versions we require in requirements.txt, it would be fine to bump the lxml version.

kalxas:
@ricardogsilva @tomkralidis it is not that simple.
There are many production deployments that have a hard requirement of deploying debian or rpm packages only: those will remain to previous versions of lxml for a long time.... some of them are still on Python 2.6... you see where this is going :)
[...]

PR #390, opened 2015-11-23
remove unused namespaces from response XML (#213)

tomkralidis:


tomkralidis:
Closing for now until lxml packages are more readily available (see https://github.com/geopython/pycsw/issues/213#issuecomment-159074283)

PR #392, opened 2015-11-27
[WIP] python 3k tests

tomkralidis:
Please do not merge.  For testing only.


tomkralidis:
cc @geopython/pycsw @QuLogic @ocefpaf

Update/status:
- Python 2.6 and 2.7 fully pass
- Python 3.4 has now progressed greatly given https://github.com/geopython/pycsw/pull/392/commits/faede63333da4ea5ba968b1163012f7f902304b0 and we are really close to fixing the remaining issues (w00t!)

Comments/questions:
- iirc we decided to go with using `six`.  Do we still need `__future__`?  Looks like there are some artifacts of us changing our minds half way through the original work

ocefpaf:
> Do we still need `__future__`? 

@QuLogic can probably elaborate this further, but the `__future__` imports (`absolute_import`, `division`, `print_function`) are not covered by using `six`. They just force the devs to abandon some Python 2 habits and code Python 3. If we drop those you'll have to write tests to ensure you are catching the proper imports, print statements, and divisions.

pavement.py
```diff
@@ -28,7 +28,12 @@
 #
 # =================================================================
 
-from ConfigParser import SafeConfigParser
+from __future__ import (absolute_import, division, print_function)
+
+try:
```
QuLogic: Should use `six.moves` here.

QuLogic:
Yes, `__future__` imports and `six` are complementary; you should remove one just because you use the other. `__future__` imports change interpreter behaviour and couldn't possibly be implemented via `six` (e.g., print statements to functions). `six` smooths out differences in the standard library, which I suppose could have been done in the interpreter, but wasn't since it's a maintenance burden.

In that vein, I'd recommend that you _always_ import the builtins from `six` to get consistent behaviour on 2 and 3: `from six.moves import filter, input, map, range, zip`.

pycsw/ogc/csw/csw3.py
```diff
@@ -2053,7 +2052,7 @@ def resolve_nsmap(self, list_):
                 prefix = tn.split(':')[0]
                 if prefix in nsmap.keys():  # get uri
                     uri = nsmap[prefix]
-                    newprefix = self.parent.context.namespaces.keys()[self.parent.context.namespaces.values().index(uri)]
+                    newprefix = list(self.parent.context.namespaces.keys())[list(self.parent.context.namespaces.values()).index(uri)]
```
QuLogic: Probably keys and values iterate in the same order, but this seems a bit fragile to me. Why not just:
`next(k for k, v in self.parent.context.namespaces.items() if v == uri)`

pavement.py
tomkralidis: Tried that and:
```python
  File "pavement.py", line 33, in <module>
    from six.moves.configparser import SafeConfigParser
```
Which I'm **guessing** is related to something in Paver..  Keeping existing unless there's a better way?

QuLogic: Maybe the higher-level `from six.moves import configparser` will work? It works fine for me on 2.7 and 3.4, so maybe it is Paver's doing.

tomkralidis: Looks like that worked, thx.

QuLogic:
Based on test results, it looks like the majority of the remaining problems are unsorted dictionary iteration.

tomkralidis:
Thanks @QuLogic.

All: in an attempt to isolate an example of the unsorted dictionary iteration issue, consider the following:

```python
d = {
    'title': '4',
    'creator': '1003',
    'subject': '29',
    'abstract': '62',
    'publisher': '1018',
    'contributor': 'TBD',
    'modified': 'TBD',
    'date': '30',
    'type': '1031',
    'format': '1034',
    'identifier': '12',
    'source': 'TBD',
    'language': 'TBD',
    'relation': 'TBD',
    'rights': 'TBD',
}

for key, value in d.items():
    print('{} => {}'.format(key, value))
```

The results are consistent between 2.6 and 2.7, but not 3.4.  Our goal is to be able to emit the above example **as specified** consistently across 2.6/2.7/3.4.

kalxas:
Interesting read: http://legacy.python.org/dev/peps/pep-0469/

kalxas:
Looking a bit deeper, it seems we will need to replace all our dict objects to collections.OrderedDict and then use list(d.items()) instead of d.items()

kalxas:
The following works both on python 2.7 and 3.4

```python
d = {
    'title': '4',
    'creator': '1003',
    'subject': '29',
    'abstract': '62',
    'publisher': '1018',
    'contributor': 'TBD',
    'modified': 'TBD',
    'date': '30',
    'type': '1031',
    'format': '1034',
    'identifier': '12',
    'source': 'TBD',
    'language': 'TBD',
    'relation': 'TBD',
    'rights': 'TBD',
}

od = collections.OrderedDict(sorted(d.items(), key=lambda t: t[0]))

for key, value in list(od.items()):
    print('{} => {}'.format(key, value))
```

tomkralidis:
This also works:

```python
from collections import OrderedDict
dict = OrderedDict

d = OrderedDict([
    ['title', '4'],
    ['creator', '1003'],
    ['subject', '29'],
    ['abstract', '62'],
    ['publisher', '1018'],
    ['contributor', 'TBD'],
    ['modified', 'TBD'],
    ['date', '30'],
    ['type', '1031'],
    ['format', '1034'],
    ['identifier', '12'],
    ['source', 'TBD'],
    ['language', 'TBD'],
    ['relation', 'TBD'],
    ['rights', 'TBD'],
])

for key, value in d.items():
    print('{} => {}'.format(key, value))
```

Note this would require some heavy lifting to change the way our dicts are **initially** defined in code ([example](https://github.com/geopython/pycsw/blob/master/pycsw/sru.py#L55-L74)), but the way they are handled thereafter is consistent/needs no change.

kalxas:
+1

kalxas:
One issue is that collections.OrderedDict is only present in Python 2.7 and 3.x, while for Python 2.6 we need the [ordereddict ](https://pypi.python.org/pypi/ordereddict) library...

QuLogic:
That might be a bit of overkill. Do you _always_ need to iterate over the dictionary in the same order? Does it need to be some order that is not the normal sort order? If it's only one time, e.g., at printing/output, and the actual order is irrelevant (as long as it's consistent),  then it's way simpler to just do:
```python
for key, value in sorted(d.items()):
   print('{} => {}'.format(key, value))
```
for the one or two times.

tomkralidis:
@QuLogic good point; we discussed on IRC today that this may be awkward/heavy.

PR #393, opened 2015-11-28
cosmetic changes to comply with pep8

ricardogsilva:
This pull request deals with cleaning up the code in `pycsw.server` in order to comply with the [official style guide for python (AKA pep8)](https://www.python.org/dev/peps/pep-0008/).

For conformance with pep8 I've used the [flake8](https://gitlab.com/pycqa/flake8) module and also the automatic checker of the Pycharm IDE.

Originally flake8 was reporting 118 issues with this module. There are now zero issues. These changes should not add or remove any functionality from the code. The intention is only to make it more readable.

My motivations:

* To bring the whole codebase closer to pep8. I've started with `pycsw.server`, but eventually other modules will also get this treatment;
* I'd like to refactor the `pycsw.server.Csw` class, specially the `dispatch` method, which is huge (roughly 300 lines of code) in order to make it 'more maintainable'. This is just a first approach, to improve readability, before the actual refactoring. 

I did not commit this directly because:

1. There is now one test that fails in the CITE suite: 

   `suites/cite/get/requests.txt:8e2232ed-05d9-44ae-8b04-0911cbe6a507`

   This test did not fail before my changes. However, it looks like this failure is due to a bug in the test itself and not because of any change introduced by this pull request. The following is the output of the test (which you should be able to replicate by running `paver test`):

   ```xml
<ows20:Exception exceptionCode="VersionNegotiationFailed" locator="acceptversions">
<ows20:ExceptionText>Invalid parameter value in acceptversions: 2006.10.29. Value MUST be 2.0.2 or 3.0.0</ows20:ExceptionText>
<ows20:ExceptionText>Invalid parameter value in acceptversions: 2006.10.29. Value MUST be 2.0.2 or 3.0.0</ows20:ExceptionText>
</ows20:Exception>
   ```
   The value for `acceptversions` seems strange. However, I'm not sure of this so I need someone to confirm if this is indeed a bug in the CITE suite or not.

2. Since the test suites probably do not cover every aspect of functionality in the affected module and since I am unable to test every corner case, I'd like more pairs of eyes to go through the changes. Most stuff is pretty inocuous but there were some `if` statements that were rewritten and I'd like to be sure that the same logic is maintained.


kalxas:
Thanks for not committing directly, we are trying to avoid that as much as possible in the Travis days :)

tomkralidis:
@ricardogsilva: awesome work here! Comments:

- echo @kalxas, thanks for the PR in lieu of direct push
- the [single test failure](https://travis-ci.org/geopython/pycsw/jobs/93679246#L1401) is rooted in [this change](https://github.com/geopython/pycsw/pull/393/files#diff-e754e02805f3a01c54bfcf3bcd61c6f2L508).  Given we have very pedantic tests, the spacing change in the response throws off the test
- I'm good with updating the expected result of this test (i.e. `mv -f tests/results/suites_cite_get_8e2232ed-05d9-44ae-8b04-0911cbe6a507.xml tests/expected`) and commit/push to this PR I'm +1 for merging

kalxas:
As long as we pass the CITE tests, I am +1 to update the expected test result and merge

ricardogsilva:
Updated the PR after @tomkralidis input. All tests are passing now. 
What is the policy now? Should I (meaning, the person who opened the PR) merge? Or should one of @tomkralidis @kalxas ?

kalxas:
we usually assign this to someone who did not commit and he merges

kalxas:
Thanks @ricardogsilva

Issue #395, opened 2016-01-13
implement JSON improvements


tomkralidis:
As per [RFC 6](http://pycsw.org/development/rfc/rfc-6.html), add better JSON support.

tomkralidis:
Added in master.

PR #396, opened 2016-01-13
implement JSON improvements (#395)

tomkralidis:
Implements [RFC6](http://pycsw.org/development/rfc/rfc-6.html).


tomkralidis:
[RFC-6](http://pycsw.org/development/rfc/rfc-6.html) now approved.

PR #406, opened 2016-02-24
Enabling Trusty build on Travis

kalxas:


kalxas:
Just to compare results...

kalxas:
ok it seems that the problem is deeper, python 2.6 works with libxml 2.9.1 and python 2.7 fails...

PR #407, opened 2016-02-25
normalize OpenSearch namespace declarations

tomkralidis:
# Overview
This PR normalizes OpenSearch namespace declaration by removing using `None`/default/implicit namespacing in lxml.  The result is having more consolidated/consistent responses across lxml versions.

# Related Issue / Discussion
None
# Additional Information
None
# Contributions and Licensing


kalxas:
Thanks for the fix!

PR #417, opened 2016-04-05
Python3 rebased

kalxas:
# Overview

# Related Issue / Discussion

# Additional Information

# Contributions and Licensing

- [X] I have already previously agreed to the pycsw Contributions and Licensing Guidelines


QuLogic:
I see no benefit to removing `__future__` even if unused.

.travis.yml
```diff
@@ -6,7 +6,7 @@ dist: trusty
 python:
   - "2.6"
   - "2.7"
-  - "3.4"
+#  - "3.4"
```
QuLogic: Instead of disabling it, add to the `known_failures` section. That way, you'll find out when it's working without having to explicitly go enable it again.

tomkralidis:
@QuLogic good point on `__future__`.  I'll merge this PR as is and let's evaluate / discuss again once the new features harden some.

Issue #427, opened 2016-04-24
Make tests more robust to small XML differences


ricardogsilva:
We are currently using a mix of `cmplib.cmp()` and `difflib.unified_diff()` for comparing test results with their expected values. This approach is flaky because XML files can have some differences in syntax but still be semantically equal, the main example being the order by which attributes are written in an XML element.

A nicer alternative to the current implementation seems to be the usage of [XML canonicalization](https://en.wikipedia.org/wiki/Canonical_XML), which is available in `lxml` by means of the `write_c14n()` method of an ElementTree object. There is a nice example of its application on this stackoverflow thread:

http://stackoverflow.com/a/30013395

tomkralidis:
:+1: Makes sense.  This would also give a bit of a performance boost, given we now have many iterations wrapped in `sorted` just to ensure tests pass, so we can then remove these given c14n would be the default means of evaluation.

With [c14n](http://lxml.de/api.html#write-c14n-on-elementtree) as the default, it would help to keep the existing in the code, which could be triggered (if desired) with a `-p` (pedantic) flag.

@ricardogsilva thoughts?

ricardogsilva:
@tomkralidis I'm preparing a PR for this and I'm facing some errors with lxml's c14n. The errors boil down to this:

```python
lxml.etree.C14NError: Relative namespace UR is invalid here : java
```

Looking into the XML file I am seeing xml namespace declarations like:

```xml
xmlns:tec="java:com.occamlab.te.TECore" 
xmlns:tems="java:com.occamlab.te.web.MonitorServlet"
```

As an example, please check `tests/suites/cite/post/f7d79701-f10b-4087-a33c-f62df0a04fd1.xml`

I guess lxml is complaining about the usage of a prefix (in this case java:) inside a namespace declaration. I don't really understand the purpose for these namespace declarations as they don't seem to be used in the rest of the XML documents.

I've looked through the [ets-csw202](https://github.com/opengeospatial/ets-csw202) CITE tests, made a search for these namespace declarations and they do not show up there. However, they are present in many (all?) POST tests of our CITE suite.

@tomkralidis do you have some insight on this? I'm stumped.

tomkralidis:
[...]

PR #435, opened 2016-05-12
fixes Issue#427

ricardogsilva:
# Overview

This Pull request makes tests more robust by comparing XML documents by using canonicalization instead of doing a simple string comparison. JSON tests are also compared by using converting python dicts to sorted lists of tuples.

# Related Issue / Discussion

More information at #427 

# Additional Information

This PR is not complete yet. I still have to merge the more recent changes in. There are also 5 tests that are failing. These are all related to the SRU suite and I believe they fail for an unrelated issue. This still needs to be clarified though.

# Contributions and Licensing


ricardogsilva:
@tomkralidis There are currently five tests that fail, but I believe them to be unrelated to this PR. They come from the fact that the prefix used in SRU requests seems to have changed recently. The test suite is expecting an `sru` prefix for the `http://www.loc.gov/zing/srw/` namespace, but the server is replying with a `zs` prefix to this namespace. Any thoughts on this?

I'll merge the recent changes to the master branch (the elapsed time stuff) into this PR and clean up a bit and it should become ready for reviewing.

tomkralidis:
Sounds good @ricardogsilva. Let's look deeper once travis is able to run.

ricardogsilva:
@tomkralidis it seems that all tests are passing in Travis now, which is a bit strange as I get 5 failed tests on my dev macchine when running under python2.7 (they all pass with python3 though). What could be the cause for the differences? Something to do with the testing database?

Another thing, I've left the older code, that performs the comparison using `cmplib.cmp`, in the file, still available by using a Pedantic flag, as @tomkralidis suggested. Personally I'm not a big fan of this. I'd rather just replace the old implementation with this one, because I'm thinking that the pedantic flag will end up not being used much (or at all). I'll let you decide.

tomkralidis:
@ricardogsilva can you squash your commits?

tests/run_tests.py
```diff
@@ -2,6 +2,7 @@
 # =================================================================
 #
 # Authors: Tom Kralidis <tomkralidis@gmail.com>
+#          Ricardo Garcia Silva <ricardo.garcia.silva@gmail.com>
 #
 # Copyright (c) 2016 Tom Kralidis
```
tomkralidis: Suggest adding your name here as well, i.e.:

```
Copyright (c) 2016 Ricardo Garcia Silva
```

ricardogsilva:
@tomkralidis I can't seem to squash cleanly. Maybe I shouldn't have made the merge with master before squashing... I've tried different ways, but I keep getting conflicts. I've got to practice rebasing more ;)
Can this go through as is?

tomkralidis:
@ricardogsilva actually GitHub gave the ability to squash and merge (nice!).  Thanks for this valuable contribution!

Issue #442, opened 2016-06-09
remove Debian setup in lieu of authoritative reference


tomkralidis:
# Description
As per @sebastic's suggestion, remove `etc/dist/debian` in lieu of `etc/dist/debian/README.txt` with a reference to https://tracker.debian.org/pkg/pycsw

@kalxas comments/thoughts?

kalxas:
I would prefer to keep some packaging files in our repo and add the README file to point to more recent packaging updates

sebastic:
Why keep the packaging in the upstream source?

It is not kept to date to not conflict with the packages included in Debian and Ubuntu causing issues when the two package sources get mixed.

Also consider the following from the [Debian UpstreamGuide](https://wiki.debian.org/UpstreamGuide#Pristine_Upstream_Source):
> Some projects include a rough /debian directory among source files to ease bleeding-edge package compilation and installation on debian (and derived) systems. While this is a good effort, it is better to leave it out of the final tarball as it can interfere with debian's own packaging effort. Keeping it only in your VCS repository is usually a much saner default if it lives in a specific packaging branch, which mimics what Debian package maintainers do using git-buildpackage. Though leaving the debian folder in the normal branch can also interfere if the package maintainer is using an upstream git packaging workflow (for example: git tag based git-buildpackage workflow). 

While there is no interference in the case of pycsw, having separately maintained packaging is generally not a good idea. It's wastes effort by duplicating work instead of collaboratively maintaining the packaging within the team that has the most expertise of Debian packaging. If you want to easily create packages for the development branches you can add another upstream branch to the Debian GIS git repository that tracks this repository instead of needing to import release tarballs, and a dedicated packaging branch (e.g. `pycsw/master`) where the pristine upstream source and Debian packaging are merged. You can easily keep that in sync with the `master` branch where the active development of the Debian package is happening (and possibly ubuntugis or osgeolive specific branches).

PR #444, opened 2016-06-25
Remove packaging

kalxas:
# Overview
Fixes issue #442 

# Related Issue / Discussion

# Additional Information

# Contributions and Licensing

- [X] I have already previously agreed to the pycsw Contributions and Licensing Guidelines


tomkralidis:
Can we add README files with info on Debian and OpenSUSE canonical locations?

kalxas:
I don't see the point to do that, we should just update the docs to inform users where to find pycsw packaged for them.

kalxas:
We are still missing a reference to DebianGIS in the installation doc, but since pycsw did not reach a stable Debian release yet, I did not add that reference yet

https://packages.debian.org/search?keywords=python-pycsw

How should we proceed? Add a reference to DebianGIS even if it points to Debian unstable?

tomkralidis:
+1 good idea.

Issue #465, opened 2016-09-23
pycsw does not decode spaces in the query part of GET requests correctly


ricardogsilva:
# Description

Whenever a GET request includes a KVP parameter whose value includes spaces that have been urlencoded to '+', pycsw is not able to correctly unquote them.

According to this [stackoverflow post](http://stackoverflow.com/questions/1634271/url-encoding-the-space-character-or-20#1634293) (which in turn references other reliable sources), the recommended way for a client to quote an HTTP GET request's **query part** is to turn spaces into the '+' character. (Note that in the **path part** of the URL, the correct quoting would be to turn spaces into '%20').  Apparently, both the '+' and the '%20' percent encoding are supported in the query part, but '+' is recommended. Either way, a robust server should be able to unquote both.

Encoding spaces as '+' is indeed how clients such as [python's requests module](http://docs.python-requests.org/en/master/) or [httpie](https://github.com/jkbrzt/httpie) work.

However, pycsw does not unquote these '+' characters back to the space character, which means that when using such clients for querying a pycsw instance, requests are not parsed correctly.

The fix for this error is pretty simple, and consists in replacing the usage of the `six.moves.urllib.parse.unquote` function with `six.moves.urllib.parse.unquote_plus`. The latter is aware of this problematic space unquoting issue and does the right thing if the request is encoded with '+' or with '%20'.
# Environment
- operating system:
- Python version:
- pycsw version:
- source/distribution
  - [x] git clone
  - [ ] DebianGIS/UbuntuGIS
  - [ ] PyPI
  - [ ] zip/tar.gz
  - [ ] other (please specify):
- web server
  - [x] Apache/mod_wsgi
  - [ ] CGI
  - [ ] other (please specify): 
# Steps to Reproduce
- Set up a local pycsw instance and use the test suite in `pycsw/tests/suites/cite` as sample data and configuration:

``` bash
# creating a docker container, to prevent messing up our own installation
docker run --rm -ti -p 8000:8000 ubuntu:16.04 /bin/bash   
apt-get update
apt-get install -y git libgeos-dev libxml2-dev libxslt1-dev libz-dev python-dev python-pip
git clone https://github.com/geopython/pycsw.git
cd pycsw
pip install -r requirements.txt -r requirements-standalone.txt -r requirements-dev.txt
pip install -e .
paver test
export PYCSW_CONFIG=${PWD}/tests/suites/cite/default.cfg
python pycsw/wsgi.py

# pycsw should be running now

```
[...]

PR #472, opened 2016-10-05
fixes issue#465

ricardogsilva:
# Overview

This commit fixes a bug with pycsw's parsing of GET parameters. As stated in the bug report, whenever the plus sign was being used to encode the space character in the query part of a URL, pycsw was not decoding the plus back into a space. This caused clients such as python's [requests](http://docs.python-requests.org/en/latest/) module to be unusable when performing for example a GetRecords request with a CQL_TEXT constraint.
# Related Issue / Discussion

This PR relates to issue #465
# Additional Information

The changes depicted in this PR are somewhat more involved than the proposed fix in issue #465. This is merely to simplify the overall code in pycsw:
- The lines removed from `pycsw.core.server` (lines 237-258) are actually pretty similar to the solution presented in python's [pep 333](https://www.python.org/dev/peps/pep-0333/#url-reconstruction). This solution has actually been implemented verbatim in python's [wsgiref.util.application_uri](https://github.com/python/cpython/blob/master/Lib/wsgiref/util.py#L43) function, which is already included in the standard library. As such it makes sense to reuse that instead of replicating it in pycsw's source code. This PR is not using wsgiref.util.application_uri directly, but rather [wsgiref.util.request_uri](https://github.com/python/cpython/blob/master/Lib/wsgiref/util.py#L63) which actually calls the other function.
- Lines 260-269 which pertain to the actual unquoting of the GET query string are replaced with the usage of some functions from `six.moves.urllib.parse`, which for python3 come from [urllib.parse](https://docs.python.org/3/library/urllib.parse.html). Among them `urllib.parse_qsl`, which does the correct decoding of the query part of the URL [by converting plus signs to spaces](https://github.com/python/cpython/blob/master/Lib/urllib/parse.py#L686) before unquoting them

The other file that has changes is `tests/suites/csw30/get/requests.txt`. This file is part of the `csw30` test suite. The modifications are the correct quoting of the values of the `outputformat` KVP parameter from

```
outputformat=application/atom+xml
```

to

```
outputformat=application/atom%2Bxml
```

Since a `+` sign in the query part of a URL is interpreted as an encoded space, the actual `+` character must be quoted as `%2B`.
# Contributions and Licensing


kalxas:
Thanks @ricardogsilva, good work!
I think we should also run the CITE tests before we merge this PR to make sure how the KVP parsing behaves there.
Thoughts?

ricardogsilva:
@kalxas sounds reasonable.

How to go about doing this official CITE testing? Should I take care of it or leave it up to you or @tomkralidis? I haven't used OGC CITE's online testing tool before, but I can learn if necessary ;)

tomkralidis:
@ricardogsilva check out the [CITE compliance wiki page](https://github.com/geopython/pycsw/wiki/OGC-CITE-Compliance) for details on how to run the CITE tests.

kalxas:
@ricardogsilva I can run the tests for you, no problem.

tomkralidis:
FYI the changes in this PR result in 1 failure against CITE 2.0.2, and skip CITE 3.0.0 tests.  I'm guessing fixing the 2.0.2 issue will fix 3.0.0 as well.

Example testcase request:

`http://host:port/?service=CSW&request=GetCapabilities&AcceptVersions=2.0.2&sections=`

in `pycsw.server.Csw.dispatch_wsgi`, dumping `self.kvp` yields:

master: `('KVP', {'AcceptVersions': '2.0.2', 'request': 'GetCapabilities', 'service': 'CSW', 'sections': ''})`

This PR: `('KVP', {'AcceptVersions': '2.0.2', 'request': 'GetCapabilities', 'service': 'CSW'})`

We see here that `sections` is set (even if empty) in master, the goal being to carry along empty CSW parameters, which are handled accordingly in later codepaths.

ricardogsilva:
@tomkralidis thanks for performing the CITE testing

This seems easy to fix, as the `parse_qsl` function accepts a `keep_blank_values` parameter that is set to `False` by default. It should be a matter of setting it to `True` in this PR's code.

However, it seems strange. [By looking at the code](https://github.com/geopython/pycsw/blob/master/pycsw/ogc/csw/csw2.py#L63), it seems that the expected behavior of `sections=` should be the same as not providing a `sections` parameter at all.

Maybe the failures in tests come from some other place? It would be handy to look at the output of the tests. I guess I'll look into running the CITE tests too and reproduce the issue locally in order to debug.

I'll report back when I get it done ;)

ricardogsilva:
@tomkralidis what is the current CITE testing URL?

I'm trying to debug the mentioned issues by performing CITE tests myself and I cannot replicate. I see all tests passing. I think I'm probably doing something wrong or following outdated instructions.

What I'm doing:

* I'm following the instructions on the [wiki](https://github.com/geopython/pycsw/wiki/OGC-CITE-Compliance#using-cite-services), 
* Created a user account at  http://cite.opengeospatial.org/teamengine/
* Installed a dummy pycsw instance in a publicly accessible endpoint using a free [pythonanywhere](https://www.pythonanywhere.com) account (their service is awesome by the way)

  * This installation is a `git clone` using as remotes both the official pycsw repo (as upstream) and my own fork (as origin) so that I can test both the current master as well as my own modifications

  * This instance is available at http://ricardogsilva.pythonanywhere.com

* Ran the CSW 2.0.2 test suite against current master. All tests pass!
* Now I've checked out this PR's branch and ran the tests again. All tests pass too!
* I don't see any CSW 3.0.0 test suite available for testing in the CITE services website.

From @tomkralidis input, it seems that somehow I should be able to test against CSW v3.0.0. What's going on here?

kalxas:
@ricardogsilva the CSW 3.0 test is on the beta team engine deployment:
http://cite.opengeospatial.org/te2/

kalxas:
and this is how to run the test locally:
https://gist.github.com/tomkralidis/58cb3ff0a383f9579f46

ricardogsilva:
@kalxas thanks for the quick feedback :)

So I've ran the tests against the online beta team engine deployment and both CSW v2.0.2 and v3.0.0 pass without any errors. I had to create a new user account and change some stuff in the `default.cfg` configuration file:

* Point `repository.database` to `<pycsw_install_dir>/tests/suites/cite/data/cite.db`
* Disable the `server.maxrecords` setting by commenting out its line
* Provide a valid e-mail in `metadata:main.contact_email` 

All of which are correctly documented in the wiki.

So I cannot replicate @tomkralidis errors. I haven't tried to run the CITE tests locally though. It seems to me that the online service should be just as reliable and perhaps even more authoritative.

@tomkralidis can you please retest this PR and see if the errors are still there? If so, can you provide some more feedback? Thanks in advance ;)

tomkralidis:
@ricardogsilva I'm still getting the same error.  I setup a pycsw instance at http://kralidis.ca:8000 by running `python pycsw/wsgi.py` along with the CITE instructions on the pycsw wiki.  I even tried the PR as a local fix against http://demo.pycsw.org/cite/csw (since reverted).

I've posted the CITE report at http://kralidis.ca/tmp/cite-csw-2.0.2-pycsw-pr-472-report.html

@kalxas any chance you can test as well?

ricardogsilva:
@tomkralidis @kalxas  I can reproduce the error now. I made a mistake in the previous test runs by forgetting to properly reload the app (since I started by testing the master branch, that was OK, later when switching to other branches and not reloading I was still testing the master always). Sorry about that.

I'll troubleshoot the failing tests and get back when they're fixed ;)

kalxas:
thank you @ricardogsilva for your continuous efforts.

ricardogsilva:
I've managed to successfully reproduce the errors and fix them. I had to make some additional adjustments, mostly to the `pycsw.opensearch` module in order to properly encode the `outputformat` parameter in the URL template from `application/atom+xml` to `application/atom%2Bxml`. Some test cases also had to be adjusted.

I've learned my CITE conformance testing 101 too ;) All tests are now passing in:

* travis
* online CITE team engine current version (CSW 2.0.2)
* online CITE team engine beta (CSW 2.0.2 and 3.0.0)

@tomkralidis or @kalxas can you please do one last round of CITE testing to verify this?

Thanks

tomkralidis:
Thanks @ricardogsilva! We should really look into hooking up the CITE tests in travis, which will allow pycsw to validate against CITE continuously, across Python versions too.

tomkralidis:
FYI see CITE ticket filed in #482

PR #474, opened 2016-10-09
Better CQL handling

tomkralidis:
For testing only.  Do not merge.


kalxas:
Looks like we have a working solution here.
@tomkralidis @ricardogsilva what do you think?

kalxas:
I am +1 to merge this PR

PR #479, opened 2016-10-19
Auto-configuration of the 'url' parameter in the config file, accordi…

mmelcot:
Ability to deploy PyCSW in different environment without to have to reconfigure manually the url parameter

### DIFF
--- a/pycsw/core/metadata.py
+++ b/pycsw/core/metadata.py
@@ -1241,9 +1241,22 @@ def _parse_iso(context, repos, exml):
             _set(context, recobj, 'pycsw:ConditionApplyingToAccessAndUse',
             md.identification.uselimitation[0])
 
-    if hasattr(md.identification, 'format'):
-        _set(context, recobj, 'pycsw:Format', md.distribution.format)
+        if hasattr(md.identification, 'format'):
+            _set(context, recobj, 'pycsw:Format', md.distribution.format)
 
+        wkt_polygon = None
+        if hasattr(md.identification, 'extent') and hasattr(md.identification.extent, 'boundingPolygon'):
+            if (hasattr(md.identification.extent.boundingPolygon, 'polygons') and
+                 len(md.identification.extent.boundingPolygon.polygons) > 0):
+                 if hasattr(md.identification.extent.boundingPolygon.polygons[0], 'exterior_ring'):
+                      exterior_ring = md.identification.extent.boundingPolygon.polygons[0].exterior_ring
+                      # Example: [(20.0, 26.0), (20.0, 35.0), (30.0, 35.0), (30.0, 25.0), (20.0, 26.0)]
+                      wkt_polygon = 'POLYGON((' + ', '.join(map(lambda x: str(x[0])+' '+str(x[1]) , exterior_ring)) + '))'
+                      # Result: POLYGON((20.0 26.0, 20.0 35.0, 30.0 35.0, 30.0 25.0, 20.0 26.0))
+                 elif hasattr(md.identification.extent.boundingPolygon.polygons[0], 'interior_rings'):
+                      # Not supported
+                      pass
+        
     if md.serviceidentification is not None:
         _set(context, recobj, 'pycsw:ServiceType', md.serviceidentification.type)
         _set(context, recobj, 'pycsw:ServiceTypeVersion', md.serviceidentification.version)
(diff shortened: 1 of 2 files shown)


pycsw/core/metadata.py
```diff
@@ -1241,9 +1241,22 @@ def _parse_iso(context, repos, exml):
             _set(context, recobj, 'pycsw:ConditionApplyingToAccessAndUse',
             md.identification.uselimitation[0])
 
-    if hasattr(md.identification, 'format'):
-        _set(context, recobj, 'pycsw:Format', md.distribution.format)
-
+        if hasattr(md.identification, 'format'):
```
ricardogsilva: The changes to `pycsw.core.metadata` are not relevant to the subject of this pull request. Therefore they should not be a part of it. Please remove them from the PR. If you wish to also propose these changes to pycsw please submit a different PR with the changes.

pycsw/wsgi.py
```diff
@@ -1,28 +1,3 @@
-# -*- coding: utf-8 -*-
```
ricardogsilva: The removal of the copyright and license header from `pycsw.wsgi` is not acceptable. Please re-add this information to the file. You may add yourself as an author, pending the acceptance of this PR.

pycsw/wsgi.py
```diff
@@ -55,18 +30,53 @@
...
+import re
+import ConfigParser
 from six.moves.urllib.parse import unquote
 
 from pycsw import server
 
 
 PYCSW_ROOT = os.path.dirname(os.path.dirname(os.path.abspath(__file__)))
 
+def dynUpdateConfigurationURLs(new_url):
```
ricardogsilva: Please modify the code to follow Python's official style guide (AKA [pep8](https://www.python.org/dev/peps/pep-0008/)). Specifically:

* use *lower_case_separated_by_underscore* for naming functions and
    variables;
* leave two blank lines before defining a new function;
* do not write lines longer than 79 characters;
* use triple double quotes for function docstrings;

pycsw/wsgi.py
```diff
@@ -55,18 +30,53 @@
...
+    '''
+    Update dynamically PyCSW configuration to match public IP:port of this server.
+
+    The PyCSW server can be deployed in a VM or a container in different environment (dev, prod...). Its IP:port is therefore not deterministic.
+    The values of 'serveraddress' in default.cfg can't be forecasted in advance.
+
+    This function is meant to be called the client emit a request.
+    The base URL requested by the client is retrieved and reinjected in PyCSW configuration.
+    '''
+    confFile = os.environ.get('PYCSW_CONFIG', '/etc/pycsw/default.cfg')
```
ricardogsilva: Reading the config file from `/etc/pycsw/default.cfg` breaks existing behaviour of reading from `<pycsw_install_dir>/default.cfg`, apparently for no good reason. This is what is causing all the [Travis CI](https://travis-ci.org/geopython/pycsw/builds/169494574) failures. Please fix this in order to read from the expected location, whenever there is no `PYCSW_CONFIG` environment variable defined.

pycsw/wsgi.py
```diff
@@ -55,18 +30,53 @@
...
+    configpar.read(confFile)
+    # Old values
+    oldServerURL = configpar.get('server', 'url')
+
+    # New values
+    new_url = re.sub('^https?://', '', new_url)  # Remove http[s] if present
+    newServerURL = 'http://' + new_url + '/pycsw/csw.py'
+
+    # Update the config is URLs are not the same as the one used for this HTTP requet.
+    if oldServerURL != newServerURL:
```
ricardogsilva: I wonder what is the point in keeping the `server.url` parameter in the configuration at all. Since the point of this PR is to make this parameter become dynamic then it seems that it should be removed from the config altogether and just used in the bits of the code that require it (mostly in the GetCapabilities operation I guess). This would negate the bug mentioned in the previous comment and actually seems like a better implementation IMHO.
It is also misleading and magical to a catalogue administrator to have a configuration parameter be changed at runtime.

tomkralidis:
There are numerous **deployment** workflows where is makes sense to hardcode `server.url` however I see the value in setting this dynamically.

Having said this, I agree with re-thinking this implementation.  Updated workflow for consideration:

1. make `server.url` an **optional** configuration
2. set `server.url` based on HTTP environment variables
3. if `server.url` is specified in configuration, then override the value set in 2.

Thoughts?  Suggest to close this PR and open as an issue for discussion and consensus first as this is a non-trivial change.

kalxas:
I am also against merging this PR, the licensing header drop is a complete show-stopper for me. 
@mmelcot thank you for your effort to improve pycsw, but as @tomkralidis said, we need a more solid solution in this case.

ricardogsilva:
Closing this since most debs are not in favor of the proposed approach and a lack of feedback from the original author. Thanks anyway @mmelcot

PR #488, opened 2016-12-01
Include dct:alternative in csw3  full output

ingenieroariel:
# Overview
We are setting up a system that configures mapproxy for CSW records. This piece of information is needed in CSW output in order to correctly configure mapproxy.

# Related Issue / Discussion

# Additional Information
Since the standard does not define exactly what goes in full, it should be okay to add this item if it is set in the database.

# Contributions and Licensing

This is needed to be able to share the WMS layer name between CSW services.


tomkralidis:
Thanks @ingenieroariel.  Backported to 2.0 branch as well.

kalxas:
Thanks @ingenieroariel

Issue #490, opened 2016-12-14
Python 3 issue with pycsw.admin


rsignell-usgs:
In a python 3 environment with pycsw 2, when I do:
```
pycsw-admin.py -c export_records -f default.cfg -p /export
```
I get: 
```
Exporting records

Processing USGS_BBLEH_SANDY_076
Writing to file /export/USGS_BBLEH_SANDY_076.xml
Traceback (most recent call last):
  File "/opt/pycsw/pycsw/core/admin.py", line 396, in export_records
    xml.write(record.xml)
TypeError: write() argument must be str, not bytes

During handling of the above exception, another exception occurred:

Traceback (most recent call last):
  File "bin/pycsw-admin.py", line 266, in <module>
    admin.export_records(CONTEXT, DATABASE, TABLE, XML_DIRPATH)
  File "/opt/pycsw/pycsw/core/admin.py", line 398, in export_records
    raise RuntimeError("Error writing to %s" % filename, err)
RuntimeError: ('Error writing to /export/USGS_BBLEH_SANDY_076.xml', TypeError('write() argument must be str, not bytes',))
```
if I change [this line in admin.py](https://github.com/geopython/pycsw/blob/master/pycsw/core/admin.py#L394) from
```
 with open(filename, 'w') as xml:
```
to
```
 with open(filename, 'wb') as xml:
```
then I get the following error:

```
Querying database sqlite:////database/cite.db, table records ....
Found 30 records

Exporting records

Processing gov.usgs.cmg:USGS_BBLEH_SANDY_076
 Sanitizing identifier
Writing to file /export/USGS_BBLEH_SANDY_076.xml
Traceback (most recent call last):
  File "/opt/pycsw/pycsw/core/admin.py", line 395, in export_records
    xml.write('<?xml version="1.0" encoding="UTF-8"?>\n')
TypeError: a bytes-like object is required, not 'str'

During handling of the above exception, another exception occurred:

Traceback (most recent call last):
  File "bin/pycsw-admin.py", line 266, in <module>
    admin.export_records(CONTEXT, DATABASE, TABLE, XML_DIRPATH)
  File "/opt/pycsw/pycsw/core/admin.py", line 398, in export_records
    raise RuntimeError("Error writing to %s" % filename, err)
RuntimeError: ('Error writing to /export/USGS_BBLEH_SANDY_076.xml', TypeError("a bytes-like object is required, not 'str'",))
```

kwilcox:
[...]

PR #491, opened 2016-12-14
Convert py3 bytes to str before exporting and test exporting

kwilcox:
# Overview

Exporting records using py3 isn't working. This could be blamed on OWSLib that can return both `bytes` and `str` for `record.xml` object. There will be a separate PR to standardize on `str` in OWSLib but to support current and older versions of `OWSLib` this fix is required.

# Related Issue / Discussion

* https://github.com/geopython/pycsw/issues/490

# Additional Information

There were no tests to handle exporting so I added them.

# Contributions and Licensing

- [x] I'd like to contribute this **py3 exporting fix** to pycsw. I confirm that my contributions to pycsw will be compatible with the pycsw license guidelines at the time of contribution.


kwilcox:
Ready for review @tomkralidis

tomkralidis:
Merged.  Thanks @kwilcox!  cc @ricardogsilva per #492

tomkralidis:
Also backported to 2.0 branch.

PR #517, opened 2017-06-01
Added a build badge to the README and also renamed it from .txt to .rst

ricardogsilva:
# Overview

This PR adds a build badge to the README in order to show off our CI state. It also renames the file from `README.txt` to `README.rst`, which seems more explicit.

# Related Issue / Discussion

See #512 

# Contributions and Licensing

(as per https://github.com/geopython/pycsw/blob/master/CONTRIBUTING.rst#contributions-and-licensing)

- [x] I'd like to contribute [feature X|bugfix Y|docs|something else] to pycsw. I confirm that my contributions to pycsw will be compatible with the pycsw license guidelines at the time of contribution.
- [x] I have already previously agreed to the pycsw Contributions and Licensing Guidelines

Issue #516, opened 2017-06-01
Store the version on a single location


ricardogsilva:
# Description

We're currently duplicating our version number on two different files:

- `VERSION.txt`
- `pycsw/__init__.py`

This is somewhat error prone, as one has to remember to keep these two in sync. Another issue is that in our `setup.py` file we are importing pycsw in order to extract the version number. This can become problematic under certain circumstances:

https://stackoverflow.com/questions/2058802/how-can-i-get-the-version-defined-in-setup-py-setuptools-in-my-package/2073599#2073599

Following the advice on the linked stackoverflow answer, we can put the version in a single file, lets say `VERSION` and then use python's `pkgresources` to extract it at runtime. This means we get to store pycsw's version in a single location.

# Environment

- operating system:
- Python version:
- pycsw version:
- source/distribution
  - [x] git clone
  - [ ] DebianGIS/UbuntuGIS
  - [ ] PyPI
  - [ ] zip/tar.gz
  - [ ] other (please specify):
- web server
  - [ ] Apache/mod_wsgi
  - [ ] CGI
  - [ ] other (please specify):

tomkralidis:
cc @kalxas 

@ricardogsilva good idea. Note when implemented/merge we should also update our Release Packaging steps (https://github.com/geopython/pycsw/wiki/Release-Packaging#step-3 and https://github.com/geopython/pycsw/wiki/Release-Packaging#step-10) accordingly.

kalxas:
+1

ricardogsilva:
I'll update the release docs too then

tomkralidis:
Just hacked them :)

PR #524, opened 2017-06-05
WIP: version is now stored on a single file, VERSION.txt

ricardogsilva:
# Overview

This PR makes `pycsw` store its version on a single location, the file `VERSION.txt`.
This change is more profound than it seems at first sight. In order to be able to have a single source of truth for the version, I had to replace `distutils` with `setuptools` - This was necessary because retrieving the installed version at runtime is done using setuptools' `pkg_resources`. Taking advantage of setuptools, I've refactored the `setup.py` file, removing code that was duplicating the job of setuptools.

However, I do not foresee a big problem with this change (lets hope the reviewers agree), as setuptools has become the standard way to install python packages and [is even included by default with python 2.7.9 and python 3.4+](https://packaging.python.org/installing/#install-pip-setuptools-and-wheel)

I'm leaving this as a WIP until PR #517 is merged because the `README` has changed extension and I'll need to update the `setup.py` code then.

# Related Issue / Discussion

See #516 

# Additional Information

# Contributions and Licensing


ricardogsilva:
It still is WIP because I'm wiating to merge #517, but please feel free to review

VERSION.txt
```diff
@@ -1 +1 @@
-2.1-dev
+2.1.0-dev
```
tomkralidis: Is this required? Our convention is not to specify the .z value when in dev, just the uneven .y.

requirements.txt
```diff
@@ -2,6 +2,7 @@ geolinks==0.2.0
 lxml==3.6.2
 OWSLib==0.13.0
 pyproj==1.9.5.1
+setuptools
```
tomkralidis: Is this required? Isn't setuptools part of the standard library?

setup.py
```diff
@@ -28,123 +30,54 @@
...
 
-def is_package(path):
-    """Decipher whether a filepath is a Python package"""
-    return (
-        os.path.isdir(path) and
-        os.path.isfile(os.path.join(path, '__init__.py'))
-    )
 
+def read(*names, **kwargs):
+    return io.open(
```
tomkralidis: Should this be wrapped in a `with` statement for file open/close safety?

setup.py
```diff
@@ -28,123 +30,54 @@
...
     maintainer='Tom Kralidis',
     maintainer_email='tomkralidis@gmail.com',
     url='http://pycsw.org/',
-    install_requires=INSTALL_REQUIRES,
-    packages=PACKAGES,
-    package_data=PACKAGE_DATA,
+    install_requires=read("requirements.txt").splitlines(),
+    packages=find_packages(),
+    include_package_data=True,
+    # TODO: replace scripts with entry_points (needs a refactoring of pycsw-admin first)
```
tomkralidis: +100!

VERSION.txt
ricardogsilva: I wasn't aware of that, sorry. I assumed we were using [semver](http://semver.org/). I'll change it.

requirements.txt
ricardogsilva: It seems it became part of standardlib [only after python 2.7.9](https://packaging.python.org/installing/#install-pip-setuptools-and-wheel), but yeah, since we've dropped py26 earlier this may not be necessary. I'll remove.

setup.py
ricardogsilva: OK I'll use `with` then.It seemed pretty innocuous not to use it here (sincewe're reading our own files) but I guess you're right ;)

ricardogsilva:
@tomkralidis I've update the code with your proposed changes.

tomkralidis:
Thanks @ricardogsilva!

ricardogsilva:
@tomkralidis this wasn't intended for merging yet (hence the WIP) ;)
I was hoping to have #517 merged in first because the extension of the `README` has changed in that PR. I'd have to adjust it here afterwards.

tomkralidis:
Oops sorry @ricardogsilva. I was too excited with these changes! Note to self to read PR titles next time :)

Issue #528, opened 2017-06-09
Running PYCSW without Environment causes configuration failure


minhd:
# Description
Using the 4 minute installation guide
```
        # Grab the pycsw source code:
        $ git clone https://github.com/geopython/pycsw.git && cd pycsw
        $ pip install -e . && pip install -r requirements-standalone.txt
        
        # Create and adjust a configuration file:
        $ cp default-sample.cfg default.cfg
        $ vi default.cfg
        # adjust paths in
        # - server.home
        # - repository.database
        # set server.url to http://localhost:8000/
        
        # Setup the database:
        $ pycsw-admin.py -c setup_db -f default.cfg
        
        # Load records by indicating a directory of XML files, use -r for recursive:
        $ pycsw-admin.py -c load_records -f default.cfg -p /path/to/xml/
        
        # Run the server:
        $ python ./pycsw/wsgi.py
        
        # See that it works!
        $ curl http://localhost:8000/?service=CSW&version=2.0.2&request=GetCapabilities
```

Results in an error
```
"GET /?service=CSW&version=2.0.2&request=GetCapabilities HTTP/1.1" 500 59
Traceback (most recent call last):
  File "/usr/lib64/python2.7/wsgiref/handlers.py", line 85, in run
    self.result = application(self.environ, self.start_response)
  File "./pycsw/wsgi.py", line 85, in application
    csw.config.get("server", "gzip_compresslevel"))
  File "/usr/lib64/python2.7/ConfigParser.py", line 607, in get
    raise NoSectionError(section)
NoSectionError: No section: 'server'
```

I've traced back the changes to the commit f8e0c88171aac52d254c4ce3ba6a7b0b4b44dbc9 where the `def get_configuration_path` is declared. 

The problem occured when there's no `PYCSW_CONFIG` global environment is declared and the `get_configuration_path` defaults to the root, which does not include `default.cfg`

I can get the server to work by explicitly declare
```
PYCSW_CONFIG=/opt/pycsw/default.cfg python ./pycsw/wsgi.py
```
and the error goes away

# Environment

- operating system:
- Python version:
- pycsw version:
- source/distribution
  - [x] git clone
  - [ ] DebianGIS/UbuntuGIS
  - [ ] PyPI
  - [ ] zip/tar.gz
  - [ ] other (please specify):
- web server
  - [x] Apache/mod_wsgi
  - [ ] CGI
  - [ ] other (please specify): 

# Steps to Reproduce
Running the 4 minute installation

[...]

PR #529, opened 2017-06-09
fix 4-minute install

tomkralidis:
# Overview
Fixes 4-minute install dev workflows
# Related Issue / Discussion
#528
# Additional Information
None
# Contributions and Licensing

### DIFF
--- a/pycsw/server.py
+++ b/pycsw/server.py
@@ -116,11 +116,10 @@ class Csw(object):
                     with codecs.open(rtconfig, encoding='utf-8') as scp:
                         self.config.readfp(scp)
         except Exception as err:
-            LOGGER.exception('Could not load user configuration: %s', err)
+            msg = 'Could not load configuration'
+            LOGGER.exception('%s %s: %s', msg, rtconfig, err)
             self.response = self.iface.exceptionreport(
-                'NoApplicableCode', 'service',
-                'Error opening configuration %s' % rtconfig
-            )
+                'NoApplicableCode', 'service', msg)
             return
 
         # set server.home safely
--- a/pycsw/wsgi.py
+++ b/pycsw/wsgi.py
@@ -92,6 +92,9 @@ def application(env, start_response):
                 "the server does not specify the 'gzip_compresslevel' option. "
                 "Returning an uncompressed response..."
             )
+        except configparser.NoSectionError:
+            print('Could not load user configuration %s' % configuration_path)
+
     start_response(status, list(headers.items()))
     return [contents]
 
@@ -163,7 +166,6 @@ def get_pycsw_root_path(process_environment, request_environment=None,
     )
     return app_root
 
-
 def get_configuration_path(process_environment, request_environment,
                            pycsw_root, config_path_key="PYCSW_CONFIG"):
     """Get the path for pycsw configuration file.
@@ -193,18 +195,23 @@ def get_configuration_path(process_environment, request_environment,
 
     """
 
+    # scan from config= or PYCSW_CONFIG environment variable
     query_string = request_environment.get("QUERY_STRING", "").lower()
+
     for kvp in query_string.split('&'):
         if "config" in kvp:
             configuration_path = unquote(kvp.split('=')[1])
             break
-    else:  # did not find any `config` parameter in the request
+    else:
+        # did not find any `config` parameter in the request
+        # lets try the process env, request env and fallback to
+        # <pycsw_root>/default.cfg
         configuration_path = process_environment.get(
             config_path_key,
-            request_environment.get(config_path_key, "")
+            request_environment.get(
+                config_path_key, os.path.join(pycsw_root, "default.cfg")
+            )
         )
-    if not os.path.isabs(configuration_path):
-        configuration_path = os.path.join(pycsw_root, configuration_path)
     return configuration_path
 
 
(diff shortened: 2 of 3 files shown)


tomkralidis:
@ricardogsilva looks like the build is failing; any ideas?

tomkralidis:
@ricardogsilva thanks for the info.  Rebased.

tomkralidis:
@ricardogsilva thanks for the review.  Rebased and re-opened for your review again.

pycsw/wsgi.py
```diff
@@ -205,6 +207,12 @@ def get_configuration_path(process_environment, request_environment,
         )
     if not os.path.isabs(configuration_path):
         configuration_path = os.path.join(pycsw_root, configuration_path)
+
+    # brute force try default.cfg in the installation directory
+    default_cfg = os.path.join(pycsw_root, 'default.cfg')
+    if os.path.isfile(default_cfg):
```
ricardogsilva: I think this test should be checking for the existence of `configuration_path` instead of `default_cfg`.

If `configuration_path` does not exist then assign `default.cfg` to `configuration_path`

tomkralidis: @ricardogsilva even in the 4-minute install case `configuration_path` indeed exists.  The above checks if `default_cfg` exists, which is an assumption that the setup has consciously created this file on disk.  Other options?

ricardogsilva: Sorry @tomkralidis maybe I am not making myself clear.

What I mean is:

- Up to line 210 we are defining the `configuration_path` variable by probing for several probable locations (`config` kvp param, `PYCSW_CONFIG` in the environment)
- Then on line 213 we check for the presence of `default_cfg`, which might hold another path different than the one found for `configuration_path`
- If `default_cfg` does exist we disregard whatever value had been previously probed from the request or the environment and just use this default. This renders the other methods of specifying the configuration pointless, especially because the `default_cfg` file is very likely to indeed exist.

In my opinion we should be testing for the existence of the path that was found by probing the request and the environment (`configuration_path`) first and use that if it exists. If this path does not exist then we'd gather the default value and use it.

Sorry for being such a PITA ;)

tomkralidis:
@ricardogsilva see latest commit (will squash once we merge).  Notes:

- I made the unit tests just work, can you review that they are ok? I'm not sure the impact of changing these unit tests per se
- slight changes in `pycsw/server.py` to make error message more safe (emitting filepath only in logs)
- catch `configparser.NoSectionError` in `pycsw/wsgi.py`.  This results in a proper CSW ExceptionReport now

It's important that we get this right given the relatively front facing functionality, so no problems continuing to improve through this PR.  Thanks again!

pycsw/wsgi.py
```diff
@@ -92,6 +92,9 @@ def application(env, start_response):
                 "the server does not specify the 'gzip_compresslevel' option. "
                 "Returning an uncompressed response..."
             )
+        except configparser.NoSectionError:
```
ricardogsilva: This `except` clause seems unnecessary here.

Since the server has been already instantiated on line [76](https://github.com/geopython/pycsw/pull/529/files#diff-27dec3b41213ad9c90fc679f954cfaf3L76) and its `__init__` method is [already catching any exceptions with trying to read the config](https://github.com/geopython/pycsw/pull/529/files#diff-e754e02805f3a01c54bfcf3bcd61c6f2R118) it seems this except clause will never be reached.

pycsw/server.py
```diff
@@ -116,11 +116,10 @@ def __init__(self, rtconfig=None, env=None, version='3.0.0'):
...
                         self.config.readfp(scp)
         except Exception as err:
-            LOGGER.exception('Could not load user configuration: %s', err)
+            msg = 'Could not load configuration'
+            LOGGER.exception('%s %s: %s', msg, rtconfig, err)
             self.response = self.iface.exceptionreport(
-                'NoApplicableCode', 'service',
-                'Error opening configuration %s' % rtconfig
-            )
+                'NoApplicableCode', 'service', msg)
```
ricardogsilva: Yes this is nice!

I also like the fact that we're not giving away the `rtconfig` value in the exception report.

pycsw/wsgi.py
```diff
@@ -193,18 +195,23 @@ def get_configuration_path(process_environment, request_environment,
...
+    else:
+        # did not find any `config` parameter in the request
+        # lets try the process env, request env and fallback to
+        # <pycsw_root>/default.cfg
         configuration_path = process_environment.get(
             config_path_key,
-            request_environment.get(config_path_key, "")
+            request_environment.get(
+                config_path_key, os.path.join(pycsw_root, "default.cfg")
+            )
```
ricardogsilva: sweet :)

tests/unittests/test_wsgi.py
```diff
@@ -66,13 +66,13 @@ def test_get_pycsw_root_path(process_env, wsgi_env, fake_dir, expected):
...
         "dummy",
-        "dummy/default.cfg"
+        "default.cfg"
     ),
     (
         {"PYCSW_CONFIG": "default.cfg"},
         {"QUERY_STRING": "config=other.cfg"},
         "dummy",
-        "dummy/other.cfg"
+        "other.cfg"
```
ricardogsilva: Seems to maintain the same spirit of the tests so I'm +1 with the adjustments.

ricardogsilva:
Thanks a lot @tomkralidis

tomkralidis:
Thanks @ricardogsilva for the careful and patient review -- much appreciated!

Issue #497, opened 2017-02-08
do not require requirements-pg.txt unless really testing PostgreSQL


tomkralidis:
# Description

Running tests seems to require a PostgreSQL installation (or at least `libpq-dev`).  For developers without PG libs installed on their system this causes errors when pip attempts to install `psycopg2` and fails with the following error message:

```
    Error: pg_config executable not found.
    
    Please add the directory containing pg_config to the PATH
    or specify the full executable path with the option:
    
        python setup.py build_ext --pg-config /path/to/pg_config build ...
    
    or with the pg_config option in 'setup.cfg'.
```

To keep things low barrier we should not require `requirements-pg.txt` when testing via the default SQLite backend.

# Environment

- operating system: Mac OS X
- Python version: 
- pycsw version:
- source/distribution
  - [x] git clone
  - [ ] DebianGIS/UbuntuGIS
  - [ ] PyPI
  - [ ] zip/tar.gz
  - [ ] other (please specify):
- web server
  - [x] Apache/mod_wsgi
  - [ ] CGI
  - [ ] other (please specify): 

# Steps to Reproduce
```bash
tox -- --exitfirst -k 'not harvesting'
```
# Additional Information

tomkralidis:
Found the problem. `requirements-dev.txt is chained to `requirements-pg.txt` (see https://github.com/geopython/pycsw/blob/master/requirements-dev.txt#L3).

Associated comment in #526 from @ricardogsilva:

>Regarding #497 I think the solution is a refactor to
>
>https://github.com/geopython/pycsw/blob/master/tests/functionaltests/conftest.py
>
>in order to load the psycopg2 module lazily, and only if needed.

@ricardogsilva: are we really able to do this in code per se (install psycopg2 lazily)?  Removing `-r requirements-pg.txt` and explicity specifing it in `tox.ini` seems like a very easy low impact solution.

PR #536, opened 2017-06-16
made psycopg2 an optional dependency for functional tests

ricardogsilva:
# Overview

This PR makes `psycopg2` an optional dependency for devs.

Currently our functional test suites are configured in the `tests/functionaltests/conftest.py` module. This module is always importing `psycopg2` regardless if the tests are using a postgres database or not.

The proposed implementation uses the [apipkg](https://github.com/pytest-dev/apipkg) package. This is a small, pure-python module that enables lazy loading of dependencies. Modules that are lazy loaded with `apipkg` are only loaded when needed. In addition to this change, the `requirements-dev.txt` file no longer includes `requirements-pg.txt`.

# Related Issue / Discussion

#497 

# Additional Information

# Contributions and Licensing

### DIFF
--- a/requirements-dev.txt
+++ b/requirements-dev.txt
@@ -1,6 +1,5 @@
 -r requirements.txt
 -r requirements-standalone.txt
--r requirements-pg.txt
 
 mock==2.0.0
 Paver==1.2.4
--- a/requirements-dev.txt
+++ b/requirements-dev.txt
@@ -1,6 +1,7 @@
 -r requirements.txt
 -r requirements-standalone.txt
 
+apipkg==1.4
 mock==2.0.0
 Paver==1.2.4
 pytest==3.0.3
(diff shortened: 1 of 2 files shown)


tomkralidis:
thanks @ricardogsilva!

sebastic:
It has been requested to revert this change in [Debian Bug #1120888](https://bugs.debian.org/1120888):
> pycsw is the very last thing using python3-apipkg
>
> python3-apipkg is not maintained & not compatible with Pytest 9.x
>
>
> Reverting this "improvement" (dubious) is all it takes to untie pycsw from python3-apipkg
>
>   https://github.com/geopython/pycsw/pull/536/files

kalxas:
Thank you @sebastic for the heads up

kalxas:
I agree with the request to revert this change, I will prepare a new PR.

PR #542, opened 2017-08-21
Filter query to show only public records

jondoig:
# Overview
Hides records that are not public (i.e. records that should be hidden from the anonymous user).  However, it also hides records with restricted access that should be accessible to the requesting user. 

Tested on Geonode 2.4 but not Geonode master.

# Related Issue / Discussion

Towards a fix for Geonode issue [#1726](https://github.com/GeoNode/geonode/issues/1726).

Works for anonymous CSW requests, but not authenticated requests.

# Contributions and Licensing

- [x] I'd like to contribute a bugfix to pycsw. I confirm that my contributions to pycsw will be compatible with the pycsw license guidelines at the time of contribution.

### DIFF
--- a/pycsw/plugins/repository/odc/odc.py
+++ b/pycsw/plugins/repository/odc/odc.py
@@ -34,6 +34,8 @@ from django.db import models
 from django.db import connection
 from django.db.models import Avg, Max, Min, Count
 from django.conf import settings
+from guardian.shortcuts import get_objects_for_user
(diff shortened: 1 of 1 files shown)


tomkralidis:
@jondoig pycsw access control is delegated to downstream applications which configure what is accessible from pycsw proper.  Work is continuing on this front in https://github.com/GeoNode/geonode/issues/1726

PR #547, opened 2017-10-30
Docker documentation update

ricardogsilva:
# Overview

This PR adds documentation on how to use pycsw's official Docker image, which has recently been merged with #534 

# Related Issue / Discussion

#530  
#534 

# Additional Information

# Contributions and Licensing

(as per https://github.com/geopython/pycsw/blob/master/CONTRIBUTING.rst#contributions-and-licensing)

- [x] I'd like to contribute [feature X|bugfix Y|docs|something else] to pycsw. I confirm that my contributions to pycsw will be compatible with the pycsw license guidelines at the time of contribution.
- [x] I have already previously agreed to the pycsw Contributions and Licensing Guidelines

PR #534, opened 2017-06-14
Adds a Dockerfile to pycsw repo

ricardogsilva:
## Overview

This PR adds a `Dockerfile` to pycsw.
This work may be continued afterwards in order to enable automated builds on docker hub.

## Related Issue / Discussion

- #530 
- [pycsw dev meeting from 2017-06-12](https://github.com/geopython/pycsw/wiki/Meetings-2017-06-12)

## Contributions and Licensing

(as per https://github.com/geopython/pycsw/blob/master/CONTRIBUTING.rst#contributions-and-licensing)

- [x] I have already previously agreed to the pycsw Contributions and Licensing Guidelines

Issue #530, opened 2017-06-09
provide a canonical Docker image


tomkralidis:
# Description
Provide a Docker setup in master which installs pycsw master with all dependencies.  Publish/configure to https://hub.docker.com/r/geopython/pycsw/

# Environment

- operating system:
- Python version:
- pycsw version:
- source/distribution
  - [x] git clone
  - [ ] DebianGIS/UbuntuGIS
  - [ ] PyPI
  - [ ] zip/tar.gz
  - [ ] other (please specify):
- web server
  - [ ] Apache/mod_wsgi
  - [ ] CGI
  - [ ] other (please specify): 

# Steps to Reproduce

# Additional Information

ricardogsilva:
@tomkralidis 
It seems that I cannot create an automated build on dockerhub, even if I am on pycsw's team. Perhaps you have to do it. Maybe I'll be able to tweak it once it is created?

Anyway, now that the PR is merged, when you have the time please create an automated build on dockerhub for pycsw. We could also do it together if you prefer

tomkralidis:
@ricardogsilva.  I deleted the project on Dockerhub and re-created via creating an automated build, which seems to have worked. I added team permissions again so you should have access to adjust build settings/etc.

ricardogsilva:
@tomkralidis thanks for that.

I still cannot do anything on docker hub though (besides seeing the repository). Must have something to do with the way docker hub assigns permissions - they are probably based on the github's repo permissions.

It would be good to configure the automated builds to:

* Build a new image with the `latest` tag whenever there is a push to `master` - This is probably already configured by dockerhub when the automated build was created
* Build a new image whenever a new tag is pushed to pycsw's repo, which will make our releases automatically available as docker images - this must be configured by us (@tomkralidis). It is fairly easy too. Just go to `Build Settings` and add a new build setting of type `Tag` (the other fields can be left as blank)

tomkralidis:
@ricardogsilva mea culpa.  Try again?

ricardogsilva:
@tomkralidis yes, I have access now. I've configured the build settings as described above. I'll proceed to add some docs on docker usage (I'll submit a PR with that) and after that we may close this issue

tomkralidis:
@ricardogsilva / @kalxas are we good to close this one?

kalxas:
+1 to close

PR #548, opened 2017-11-05
Dockerfile improvements

ricardogsilva:
# Overview

This PR improves the existing `Dockerfile` by adding a few missing stuff from the first implementation. Namely:

- Addition of a `--reload` flag that makes the web server automatically reload whenever the code changes. This is only useful for development purposes, as it simplifies the dev workflow;
- Usage of python's `os.execlp` function, which behaves similarly to bash's `exec`. This causes the entrypoint script to be replaced by the web server (in this case gunicorn) in such a way that the same environment and PID is used. This effectively means that gunicorn is now PID1 of the docker container, making it easier to interact with it. It also means that a running pycsw container will now exit cleanly without errors when it is stopped or removed.

# Related Issue / Discussion
#530  
#534

# Additional Information

I have another also PR #547, which is related to this one. It adds documentation on how to use the docker image.

# Contributions and Licensing

### DIFF
--- a/requirements-dev.txt
+++ b/requirements-dev.txt
@@ -8,5 +8,4 @@ pytest==3.0.3
 pytest-cov==2.4.0
 pytest-flake8==0.8.1
 pytest-timeout==1.2.0
-requests==2.10.0
 sphinx
--- a/docker/entrypoint.py
+++ b/docker/entrypoint.py
@@ -43,7 +43,7 @@ import logging
 import os
 from six.moves.configparser import SafeConfigParser
 from six.moves.configparser import NoOptionError
-from subprocess import call
+import sys
 from time import sleep
 
 from sqlalchemy import create_engine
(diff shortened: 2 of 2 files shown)


ricardogsilva:
It seems I've messed up this PR with the documentation one #547. Sorry about that. 
Let me try to clean things up.

ricardogsilva:
OK, all good now. Please review at your convenience.

tomkralidis:
+1

requirements-dev.txt
```diff
@@ -8,5 +8,4 @@ pytest==3.0.3
 pytest-cov==2.4.0
 pytest-flake8==0.8.1
 pytest-timeout==1.2.0
-requests==2.10.0
```
kalxas: is this removal related to this PR?

ricardogsilva: Actually (and strangely) yes it is.

The `requests` dependency does not seem to be needed for pycsw - I've done some grepp'ing in the codebase and I don't think we are using it anywhere (I might be wrong though).
 
In addition to that (and hence why it is related here), the gunicorn wsgi server, which is being used in the docker image, depends on a more recent version of requests. 
When I was testing the workflow for getting a development environment based on docker (as described in the docs on the other PR) I realized that when one would do 

```
pip install -r requirements-dev.txt
```

inside a running container it would uninstall the version of requests required by gunicorn and downgrade it to the one pinned in our requirements-dev.txt. This would lead to a crash in gunicorn when reloading/restarting the server's worker processes.

kalxas: thanks!

PR #555, opened 2017-11-12
update requirements versions, add Python 3.7 to Travis

tomkralidis:
# Overview
- unpegs requirements versions
- adds 3.7 to Travis testing
- drops 2.7 testing as a Travis `allow_failures`

# Related Issue / Discussion

# Additional Information

# Contributions and Licensing


tomkralidis:
@ricardogsilva any advice on how to fix the test failures?  They work on my env (2.7), but I can't see what the errors actually are on Travis.

.travis.yml
```diff
@@ -2,7 +2,7 @@ language: python
 
 cache: pip
 
-dist: trusty
+dist: xenial
```
ricardogsilva: AFAIK travis does not have a `xenial` image. According to their [docs](https://docs.travis-ci.com/user/reference/overview/) only precise and trusty are available. Also the [test's logs](https://travis-ci.org/geopython/pycsw/jobs/306947343#L9) show that in fact trusty is being used and not xenial. So this should be reverted back to trusty` it seems.

.travis.yml
```diff
@@ -32,21 +32,21 @@ jobs:
       python: "2.7"
       env: TOXENV=py27-sqlite
       install: pip install codecov tox
-      script: tox -- --exitfirst -m functional -k 'not harvesting'
+      script: tox -- --exitfirst -m functional -k 'not harvesting' --functional-prefer-diffs
```
ricardogsilva: Didn't we agree that the `--functional-prefer-diffs` flag was not to become the default behaviour for tests? If so I don't think it should be set on Travis either, since these are our default CI tests.

ricardogsilva:
@tomkralidis 
I did some initial investigation on this. Some remarks:

- I also get these errors on my local env (which is a docker container, like the one described in [pycsw's docs](http://docs.pycsw.org/en/latest/docker.html#setting-up-a-development-environment-with-docker)
- Strangely the diff printing does not show anything -  I did not dig into this yet
- When I run the tests without the `--functional-prefer-diffs` flag this concrete test is actually not failing. There are errors in other tests though;
- From what I could gather it seems that errors show up in tests that are verifying exceptions. I'd say that probably lxml or some other dependency that you're trying to update with this PR is making the exception text be different now and that is throwing the errors. This goes to show that the functional tests are perhaps too brittle. IMHO we should not be comparing things so closely (i.e. matching the exact text that is shown after an exception is raised), but rather do a more semantic comparison (in this case just verifying that the response is an ExceptionReport with the correct code and locator would be enough). I understand that what I am proposing would require a big refactoring of the functional tests and is not feasible for now though
- Running all tests (except harvesting) without the `--functional-prefer-diffs` flag I get 431 passes and 4 fails.
- When I run them with `--functional-prefer-diffs` I get 428 passes and 7 fails.
- It seems to me that we should not set `--functional-prefer-diffs` flag on Travis. From our previous discussion I understood that this flag was not to be the default behaviour and in accordance to that it should also not be enabled on our main CI tests I guess. I also consider XML canonicalization to be a superior method of comparison in this case.

I can try to dig a bit on the wierd behaviour of not showing any difference in the specific test that is failing but I admit I'm not too familiar with how python's `difflib` works so perhaps you might also lend a hand with this?

In a more pragmatic approach, it seems to me that in order to fix the missing tests one would need to:

- drop the `--functional-prefer-diffs` flag from travis
- hunt down the exception messages of the failing tests, compare them with the expected values and update as needed

.travis.yml
tomkralidis: @ricardogsilva agree/+1.  I turned these on temporarily to try to see what the error was (my local built fine).

.travis.yml
tomkralidis: Thanks @ricardogsilva; reverted.

tomkralidis:
Update: I've done the following:

- unpegged versions in `requirements.txt`
- update Travis build to xenial
- added Python 3.7

Everything works/passes except for 2.7.  Errors are rooted in different lxml error messages between Python 2.7 and 3.x.  I've emailed the lxml mailing list for guidance.

If we are still stuck, the following options exist:
- emit deeper error messages only to logs and keep error messages short/consistent to pycsw messaging (i.e. "Invalid XML document")
- drop support for 2.7 testing in Travis
- something else?

.travis.yml
tomkralidis: Update: Travis now has a xenial image, reverted back.  This helps with Python 3.7 testing.

kalxas:
The question is if tests are still working in Python 2.7
I would not mind dropping Travis + Python 2.7 but I would hesitate dropping Python 2.7 completely.

tomkralidis:
We can stop testing Python 2.7 in Travis.  Note we would **not** remove any code which may be Python 2.7 related.

Another option is to use Travis' `allow_failures` option which would always tests on 2.7, but do not count in the assessment of the build.  This would allow us to keep testing 2.7 and only inspect when needed.

tomkralidis:
I tested `allow_failures` and this seems to work: https://travis-ci.org/geopython/pycsw/builds/501641999

Are folks ok with this approach?  If yes I'll clean up this PR.

kalxas:
I am ok with this

PR #564, opened 2018-03-22
Use xdg-open rather than using firefox explicitely

johanvdw:
xdg-open will use the users prefered browser

# Overview

# Related Issue / Discussion

# Additional Information

# Contributions and Licensing


johanvdw:
Note this patch was pas part of the debian package for a while

kalxas:
Thanks @johanvdw

Issue #561, opened 2018-03-06
Harvest fails with Python 3


juozasg:
# Description
Trying to harvest from remote CSW using a POST call fails with error:  `<ows:ExceptionText>Harvest failed: record parsing failed: 'bytes' object has no attribute 'tag'</ows:ExceptionText>`

This is probably because `util.http_request` returns a `bytes` object. But `metadata._parse_metadata` checks only for `str` type.

# Environment
- operating system: OS X (Homebrew)
- Python version: 3.6.4
- pycsw version: 2.0.3
- source/distribution
  - [x] PyPI
- web server
  - [x] other (please specify): python3 pycsw/wsgi.py 

# Steps to Reproduce
curl -X POST --data-binary @sync-POST.xml 'http://127.0.0.1:8000'

```
<?xml version="1.0" encoding="UTF-8"?>
<csw:Harvest xmlns:csw="http://www.opengis.net/cat/csw/2.0.2" xmlns:gmd="http://www.isotc211.org/2005/gmd" service="CSW" version="2.0.2">
  <csw:Source>https://rda.ucar.edu/geoportal/csw/discovery?Request=GetCapabilities&amp;Service=CSW&amp;Version=2.0.2</csw:Source>
  <csw:ResourceType>http://www.isotc211.org/schemas/2005/gmd/</csw:ResourceType>
</csw:Harvest>
```

# Additional Information

eblondel:
If it helps for reproducing: I got exactly the same error using docker geopython/pycsw latest image (run on OS Win 10) and using the test request https://github.com/geopython/pycsw/blob/master/tests/functionaltests/suites/default/post/Harvest-default.xml

PR #574, opened 2018-08-22
Support indexing the ISO files from the IOOS Catalog

kwilcox:
# Overview
I was attempting to setup a `pycsw` server for IOOS and ran across a few hiccups that I fixed in this PR. I also fixed an outstanding issue (#561) that I ran into when running the functional tests.

# Related Issue / Discussion
In general: it was not easy to debug loading a very large folder of ISO files into `pycsw`. The loops would error out if a file couldn't be parsed and the error messages were not specific enough to debug which files were cause the issues. In some cases the error messages were much to verbose because they contained the entire SQL INSERT statement. The logging when loading records is much more user friendly now. While making the logging more friendly I came across a few other hiccups that I fixed along the way.

# Additional Information

# Contributions and Licensing

- [x] I'd like to contribute these commits to pycsw. I confirm that my contributions to pycsw will be compatible with the pycsw license guidelines at the time of contribution.


kwilcox:
I'm running the tests against sqlite and postgres to make sure the VACUUM fix is actually fixed...

kwilcox:
@tomkralidis ready for review

pycsw/core/repository.py
```diff
@@ -300,9 +300,7 @@ def insert(self, record, source, insert_date):
             self.session.commit()
         except Exception as err:
             self.session.rollback()
-            msg = 'Cannot commit to repository'
-            LOGGER.exception(msg)
-            raise RuntimeError(msg)
+            raise
```
tomkralidis: Will (only) `raise` emit `msg`?

kwilcox: Yup, this just re-raises the same exception that was caught in the `except` block. This was used so I could catch DB errors in the `admin.load_records` function and return a specific error rather than a generic message/exception.

PR #580, opened 2018-12-20
Fix #576:  Error when querying with ISO output schema

### LINKED ISSUE #576: Error when querying with ISO output schema
# Description
When using `GetRecordByID` request with `outputSchema` parameter set to `http://www.isotc211.org/2005/gmd` the server responds with a 500 error. When the `outputSchema` is not set, it returns the record in default schema with no problems.

# Environment

- operating system: CentOS 7.5.1804
- Python version: 3.4.9
- pycsw version: 2.2.0
- source/distribution
  - [ ] git clone
  - [ ] DebianGIS/UbuntuGIS
  - [x] PyPI
  - [ ] zip/tar.gz
  - [ ] other (please specify):
- web server
  - [ ] Apache/mod_wsgi
  - [ ] CGI
  - [x] other (please specify): Gunicorn

# Steps to Reproduce

Example request:
```
GET /?service=CSW&version=2.0.2&request=GetRecordById&elementsetname=full&outputSchema=http:%2F%2Fwww.isotc211.org%2F2005%2Fgmd&id=12345
```

# Additional Information

Inspecting the logs reveals a python TypeError:
```
File "/venvs/pycsw_venv/lib/python3.4/site-packages/pycsw/plugins/profiles/apiso/apiso.py", line 389, in write_record
    if caps is None and xml_blob is not None and xml_blob.startswith(b'<gmd:MD_Metadata'):
TypeError: startswith first arg must be str or a tuple of str, not bytes
```
To get around this I've made adjustments in our running instance of `pycsw/plugins/profiles/apiso/apiso.py`.

Current master version (lines 388-394):
```python
xml_blob = util.getqattr(result, self.context.md_core_model['mappings']['pycsw:XML'])
if caps is None and xml_blob is not None and xml_blob.startswith(b'<gmd:MD_Metadata'):
    is_iso_anyway = True

if (esn == 'full' and (typename == 'gmd:MD_Metadata' or is_iso_anyway)):
    # dump record as is and exit
    return etree.fromstring(xml_blob, self.context.parser)
```

Adjusted version:
```python
xml_blob = util.getqattr(result, self.context.md_core_model['mappings']['pycsw:XML'])
xml_blob_bytes_object = bytes.fromhex(xml_blob[2:]) # Convert long hex value to bytes object (trim \x from start of string so that fromhex works)
xml_blob_decoded = xml_blob_bytes_object.decode('utf-8') # Then decode the bytes to unicode string

if caps is None and xml_blob is not None and xml_blob_decoded.startswith('<gmd:MD_Metadata'): # Removed `b` prefix, as it is checking a str. Also check on new `xml_blob_decoded` variable instead
    is_iso_anyway = True

if (esn == 'full' and (typename == 'gmd:MD_Metadata' or is_iso_anyway)):
    # dump record as is and exit
    return etree.fromstring(xml_blob_decoded, self.context.parser) # use new `xml_blob_decoded` variable that stores real unicode string value of the XML.
```

I get the real XML markup (not hex as per database value) and store in `xml_blob_decoded` because `fromstring` expects to parse real XML markup. I'm unsure if this is the correct way to patch this, but it now allows me to search for a record with ISO output schema at least as a temporary workaround.

juozasg:
This fixes a Python 3 error when querying ISO gmd:MD_Metadata

The fix is provided by @ludona7 in this ticket: https://github.com/geopython/pycsw/issues/576

### DIFF
--- a/pycsw/plugins/profiles/apiso/apiso.py
+++ b/pycsw/plugins/profiles/apiso/apiso.py
@@ -386,12 +386,15 @@ class APISO(profile.Profile):
         is_iso_anyway = False
 
         xml_blob = util.getqattr(result, self.context.md_core_model['mappings']['pycsw:XML'])
-        if caps is None and xml_blob is not None and xml_blob.startswith(b'<gmd:MD_Metadata'):
+        xml_blob_bytes_object = bytes.fromhex(xml_blob[2:]) # Convert long hex value to bytes object (trim \x from start of string so that fromhex works)
+        xml_blob_decoded = xml_blob_bytes_object.decode('utf-8') # Then decode the bytes to unicode string
+
+        if caps is None and xml_blob is not None and xml_blob_decoded.startswith('<gmd:MD_Metadata'): # Removed `b` prefix, as it is checking a str. Also check on new `xml_blob_decoded` variable instead
             is_iso_anyway = True
 
         if (esn == 'full' and (typename == 'gmd:MD_Metadata' or is_iso_anyway)):
             # dump record as is and exit
-            return etree.fromstring(xml_blob, self.context.parser)
+            return etree.fromstring(xml_blob_decoded, self.context.parser) # use new `xml_blob_decoded` variable that stores real unicode string value of the XML.
 
         node = etree.Element(util.nspath_eval('gmd:MD_Metadata', self.namespaces))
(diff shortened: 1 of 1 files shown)


tomkralidis:
Closing in lieu of #596

PR #587, opened 2019-02-17
fix to handle '/' or '\\'

ahmednoureldeen:
# Overview
simply use os.sep instead of '/' to handle Windows or Linux  systems

# Related Issue / Discussion
this fixes issue in Windows
Could not load custom mappings: Import by filename is not supported.
Traceback (most recent call last):
  File "D:\work\geonode\env_10\lib\site-packages\pycsw\server.py", line 192, in __init__
    mappings = __import__(module, fromlist=[''])
ImportError: Import by filename is not supported.

# Contributions and Licensing

- [ *] I'd like to contribute bugfix  to pycsw. I confirm that my contributions to pycsw will be compatible with the pycsw license guidelines at the time of contribution.
- [* ] I have already previously agreed to the pycsw Contributions and Licensing Guidelines

### DIFF
--- a/pycsw/server.py
+++ b/pycsw/server.py
@@ -184,7 +184,7 @@ class Csw(object):
             try:
                 import imp
                 module = self.config.get('repository', 'mappings')
-                if '/' in module:  # filepath
+                if os.sep in module:  # filepath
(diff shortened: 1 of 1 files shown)


tomkralidis:
Thanks @ahmednoureldeen !

PR #598, opened 2019-04-26
Fix #597 Handle different formats for CRS code input

### LINKED ISSUE #597: load_records duplicate CRS value for ISO
# Description
I came across an issue with the CRS data when using the [`load_records`](https://docs.pycsw.org/en/1.8.1/administration.html#loading-records) method, specifically when importing ISO records.

Once imported, I found that my database stored values in `records.crs` like this:

`urn:ogc:def:crs:EPSG:6.11:urn:ogc:def:crs:EPSG::4326`

Inspecting the code in https://github.com/geopython/pycsw/blob/master/pycsw/core/metadata.py#L1245, it prepends `urn:ogc:def:crs:EPSG:6.11` to the value of `referenceSystemIdentifier.RS_Identifier.code` of the source XML file.

This works in cases where just the 'number code' is defined in `referenceSystemIdentifier.RS_Identifier.code` (e.g. 4326), but I found that some of the records I wanted to import (rightly or wrongly) have the whole value like: `urn:ogc:def:crs:EPSG:6.11:4326` sometimes with or without a version.

I've created a patch on my instance that should now cater for all the possible formats it might receive when it loads the records into pycsw - (ie. just the num code, the full string with or without a version).

I've also removed the hardcoded 'version' (6.11) that was being prepended, and replaced with the extracted 'version' from the value of `<gmd:code>`, or if `<gmd:version>` is defined, it takes it from here instead.

I will create a PR in case you think this is useful more generally.

# Environment

- operating system: CentOS 7.5.1804
- Python version: 3.4.9
- pycsw version: 2.2.0
- source/distribution
  - [ ] git clone
  - [ ] DebianGIS/UbuntuGIS
  - [x] PyPI
  - [ ] zip/tar.gz
  - [ ] other (please specify):
- web server
  - [ ] Apache/mod_wsgi
  - [ ] CGI
  - [x] other (please specify): nginx/gunicorn

# Steps to Reproduce

Run `load_records` function where the input XML file is ISO and contains a 'full' reference in the `referenceSystemIdentifier.RS_Identifier.code` element, e.g `urn:ogc:def:crs:EPSG::4326`

# Additional Information

ludona7:
# Overview

This attempts to fix an issue with CRS duplication when records loaded in via `load_records`.

# Related Issue / Discussion

#597 

# Additional Information

n/a

# Contributions and Licensing


tomkralidis:
Can this PR (or a new one) be updated per https://github.com/geopython/pycsw/issues/597#issuecomment-494030543 ?

Issue #614, opened 2020-05-07
optimize Dockerfile


tomkralidis:
# Description
cc @epifanio @ricardogsilva 

Carrying on from https://github.com/geopython/pycsw/pull/613#pullrequestreview-407505149, find ways to speed up build, reduce image size, and testing.

# Environment

- operating system:
- Python version:
- pycsw version:
- source/distribution
  - [ ] git clone
  - [ ] DebianGIS/UbuntuGIS
  - [ ] PyPI
  - [ ] zip/tar.gz
  - [ ] other (please specify):
- web server
  - [ ] Apache/mod_wsgi
  - [ ] CGI
  - [ ] other (please specify): 

# Steps to Reproduce
NA

# Additional Information
NA

tomkralidis:
@ricardogsilva the following test builds in 40 seconds but creates an image of 334MB.  Any ideas to reduce the size?

```
FROM python:slim-buster
LABEL maintainer="massimods@met.no,aheimsbakk@met.no"

ARG PYCSW_HOME=/tmp/pycsw

ENV PYCSW_CONFIG=/etc/pycsw/pycsw.cfg

COPY . ${PYCSW_HOME}

RUN apt-get update && apt-get install -y ca-certificates

RUN pip3 install --upgrade pip setuptools \
  && pip3 install wheel \
  && pip3 install gunicorn \
  && pip3 install --requirement ${PYCSW_HOME}/requirements.txt \
  && pip3 install --requirement ${PYCSW_HOME}/requirements-standalone.txt \
  && pip3 install --requirement ${PYCSW_HOME}/requirements-pg.txt

ADD docker/pycsw.cfg ${PYCSW_CONFIG}
ADD docker/entrypoint.py /usr/local/bin/entrypoint.py

WORKDIR ${PYCSW_HOME}

RUN pip3 install . \
  && useradd -ms /bin/bash pycsw \
  && cp -r ${PYCSW_HOME}/tests /home/pycsw \
  && chown -R pycsw:pycsw /home/pycsw/* \
  && rm -rf /usr/lib/python3*/*/tests \
  && rm -rf /usr/lib/python3*/ensurepip \
  && rm -rf /usr/lib/python3*/idlelib \
  && rm -f /usr/lib/python3*/distutils/command/*exe \
  && rm -rf /usr/share/man/* \
  && rm -fr ${PYCSW_HOME} \
  && find /usr/lib -name  "*.pyc" -o -name "*.pyo" -delete

[...]

PR #615, opened 2020-05-10
refactor Dockerfile in order to use python:3.8-slim-buster as base

ricardogsilva:
# Overview
This PR modifies pycsw's `Dockerfile` in order to use an official python-debian image as the base for building.

# Related Issue / Discussion

#614

# Additional Information

My percieved advantages of this approach vs the current implementation (base image based on alpine linux):

-  Faster cold build time, since we can use python wheels - no need to compile anything
-  Similar environment to CI and non-docker environments (seems to me that alpine is not common outside of embedded and containers)
-  Being based on the official Python repos, it is easier to build and test pycsw with other python versions, it is just a matter of changing the dockerfile base image

In addition to replacing alpine with an offcial python image based on current debian, I've made some other notorious modifications to the `Dockerfile`:

-  Rearrange dockerfile commands in order to make use of the docker caching mechanism. This makes rebuilding the image much faster whenever a change is made in the pycsw codebase - This change is orthogonal to the image being used as a base so we should add it regardless of what final decision is made on the alpine vs. debian front

-  Install pycsw in editable mode, which allows easier dev workflow - we can run a container and mount the code dir

   Additionally, leave all files in place instead of moving pycsw config to `/etc/pycsw` and moving `entrypoint.py` to `/usr/local/bin` - this is also in order to allow a friendlier dev environment

-  Install the `psycopg2-binary` dependency instead of `psycopg2`. This change has been made in the `Dockerfile` and not on the `requirements-pg.txt` file in order to be less intrusive

Comparison between the proposed implementation and current master:

-  Current master (based on alpine)

   -  Builds are slow
   -  image is small
   -  Rebuilds are slow

   ```
   time docker build -t geopython/pycsw:master -f Dockerfile .
   # build time: 2m50s
   # image size: 172MB
   # rebuild time after a change in pycsw's codebase: 2m58s
   ```

-  This implementation (based on debian)

   -  Builds are fast
   -  image is bigger
   -  Rebuilds are fast

   ```
   docker rmi geopython/pycsw:with-dev python:3.8-slim-buster
   time docker build -t geopython/pycsw:without-dev -f Dockerfile.test2 .
   # build time: 30s
   # image size: 311MB
   # rebuild time after a change in pycsw's codebase: 5s
   ```

IMO an image size of 311MB is still pretty modest and, all things considered, [absolute image size is probably not so relevant anyway](https://semaphoreci.com/blog/2018/03/14/docker-image-size.html).

This implementation builds significantly faster and uses a more flexible and familiar base image. It also opens up the door to other niceties:

-  It is now easier to dev inside a docker container, for example with:

   ```
   # using the container for dev work by bind mounting the code dir inside the container
   # changes to pycsw are automatically available inside the container
   docker run \
       --name pycsw-dev \
       --rm \
       -ti \
       -v $(pwd)/pycsw:/home/pycsw/pycsw/pycsw \
       {image}

   # reload the gunicorn server by sending the HUP signal (it is pid 1 inside the container)
   docker exec -ti pycsw-dev /bin/bash -c 'kill -HUP 1'
   ```

-  It is now easier to run tests inside a docker container:

   ```
   # running tests inside container (requires installation of the dev requirements)
   docker run \
       --rm \
       -ti \
       --user root \
       --entrypoint /bin/bash \
       {image} \
       -c 'cd pycsw && pip install requirements-dev.txt && su pycsw -c "pytest -x -v -k \"not harvesting and not distributed\""'
   ```

# Contributions and Licensing

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -5,7 +5,7 @@
 # Contributors: Arnulf Heimsbakk <aheimsbakk@met.no>
 #               Tom Kralidis <tomkralidis@gmail.com>
 #
-# Copyright (c) 2017 Ricardo Garcia Silva
+# Copyright (c) 2020 Ricardo Garcia Silva
 # Copyright (c) 2020 Massimo Di Stefano
 #
 #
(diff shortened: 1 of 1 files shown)


ricardogsilva:
There seems to be a failing CI test for something unrelated to this. Something to do with demo.pycsw.org

@tomkralidis any thoughts on this?

tomkralidis:
@ricardogsilva I revamped the [demo](https://github.com/geopython/demo.pycsw.org) and forgot that the gisdata service was being used by the tests.  This is back online now, and I've closed/reopened this PR as a result to see how the tests run.

ricardogsilva:
@tomkralidis any update on this?

Dockerfile
```diff
@@ -33,66 +33,40 @@
 #
 # =================================================================
 
-FROM alpine:3.11
+FROM python:3.8-slim-buster
 LABEL maintainer="massimods@met.no,aheimsbakk@met.no"
 
-ARG PYCSW_HOME=/tmp/pycsw
+RUN apt-get update && apt-get install --yes \
```
tomkralidis, line 39: Should we add `--no-install-recommends` here (recommended by https://www.fromlatest.io) ?

tomkralidis:
@ricardogsilva in discussion : https://gitter.im/geopython/pycsw?at=5ebb568840e3122a8814ef35

I like the fact that the PR aligns the dev and test workflow.  cc'ing @kalxas and @epifanio for any comments on this PR.

kalxas (approved):

kalxas:
Thanks for the updates.
+1 to merge

kalxas:
@ricardogsilva @epifanio @tomkralidis how about having 2 separate docker images (one based on alpine and one based on debian)?

tomkralidis:
+1

ricardogsilva:
-1

I'm obviously biased, since I am the one proposing this change :smile: 

However, it seems to me that the only benefit of keeping an alpine-based image at this point is just to try and get the smallest image size possible. In my opinion this is not a worthy effort because:

- the current proposal generates images that are still somewhat small (for Python), as it is below 350 MB
- due to how docker caching works, the actual size taken on disk is likely much smaller, since deployments are likely to also use other images based on debian
- being based on glibc, we can use python wheels (which are not available for alpine) and thus we don't need to install dev dependencies no compile anything, which results in a simpler Dockerfile
- build times are a lot faster and rebuilds are also faster
- there is now a much more pleasant developer experience
- opens up interesting possibilities also from a CI perspective, where we could now have a pipeline that builds the docker image, runs tests inside it, pushes it to docker registry, etc

Also, keeping two images means more maintenance work, which is also a stress factor given the small size of the team.

kalxas:
I am also biased because I do much packaging for rpm and deb so I obviously prefer a mature distribution over alpine :)

epifanio:
I am also biased, but I will try to give my 2 cents. 
At work I was required to use the smallest image possible ... hence the needs of using alpine. In reality, as a developer, the benefits of using an image with a small footprint like alpine are too little compared to the flexibility offered by Debian based machine. 
Maybe in a production environment alpine is to prefer and will be worth to maintain a pycsw dockerfile dedicated to that purpose ... but in development, I found very time-consuming rebuilding each time the alpine machine because I needed to add extra libraries for data-science which are not available for alpine (also debugging is hard, no vim no nano no bash .. in default alpine)

kalxas:
Thank you all for the feedback, I think there is consensus to merge this PR and move to debian based images as default.

PR #622, opened 2020-08-06
test invalid gml:posList geometry (#621)

### LINKED ISSUE #621: order of coordinates in posList
I have tried to set up a spatial query for a CSW server, once with a `gml:Point` (````<gml:pos srsDimension='2'>48.1537 16.2825 </gml:pos>````) and once with a `gml:Polygon` (e.g., ```<gml:posList srsDimension='2'>48.1537 16.2825 48.1537 16.3759  48.2201 16.3759 48.2201 16.2825 48.1537 16.2825</gml:posList>```) geometry. I have assigned the same spatial reference system to both (EPSG:4326), which prescribes the order of the points to be latitude first, longitude second. However, the returned results differ completely in space, where the point query was right, but the polygon query was wrong. 

Having a look at the code, which parses GML in pyCSW it turns out that `gml:Point/gml:pos` are differently parsed than `gml:Polygons/gml:posList`. The following lines seem wrong to me:
https://github.com/geopython/pycsw/blob/b5a657e1e0bc7fbfb327d5635d4bdfc9d15c6f5d/pycsw/ogc/gml/gml3.py#L50-L51.
Assuming `axisorder='xy'` points are parsed latitude first, longitude second, which is okay. But in the two lines above `xlist` are the longitudes and `ylist` the latitudes, which is wrong. The following should solve the problem:
```
xlist = tmp[::2]
ylist = tmp[1::2]
```
No matter what geometry parsing is modified, at least they should be consistent. I hope I did not oversee something or got this wrong, thanks!

tomkralidis:
# Overview
Adds testing `gml:posList` length of geometry tuples.

# Related Issue / Discussion
Fixes #621 
# Additional Information
NA
# Contributions and Licensing

### DIFF
--- a/pycsw/ogc/gml/gml3.py
+++ b/pycsw/ogc/gml/gml3.py
@@ -41,14 +41,24 @@ TYPES = ['gml:Point', 'gml:LineString', 'gml:Polygon', 'gml:Envelope']
 DEFAULT_SRS = crs.Crs('urn:x-ogc:def:crs:EPSG:6.11:4326')
 
 
-def _poslist2wkt(poslist, axisorder):
+def _poslist2wkt(poslist, axisorder, geomtype):
     """Repurpose gml:posList into WKT aware list"""
 
     tmp = poslist.split()
     poslist2 = []
 
-    xlist = tmp[1::2]
-    ylist = tmp[::2]
+    if geomtype == 'polygon':
+       len_ = 8
+    elif geomtype == 'line':
+       len_ = 4
+
+    if len(tmp) < len_:
+        msg = 'Invalid number of coordinates in geometry'
+        LOGGER.error(msg)
+        raise RuntimeError(msg)
+
+    xlist = tmp[::2]
+    ylist = tmp[1::2]
 
     if axisorder == 'yx':
         for i, j in zip(ylist, xlist):
@@ -134,7 +144,8 @@ class Geometry(object):
                                Missing gml:posList')
         else:
             self.wkt = 'LINESTRING(%s)' % _poslist2wkt(tmp.text,
-                                                       self.crs.axisorder)
+                                                       self.crs.axisorder,
+                                                       'line')
 
     def _get_polygon(self):
         """Parse gml:Polygon"""
@@ -147,7 +158,8 @@ class Geometry(object):
                                Missing gml:posList')
         else:
             self.wkt = 'POLYGON((%s))' % _poslist2wkt(tmp.text,
-                                                      self.crs.axisorder)
+                                                      self.crs.axisorder,
+                                                      'polygon')
 
     def _get_envelope(self):
         """Parse gml:Envelope"""
--- a/pycsw/core/repository.py
+++ b/pycsw/core/repository.py
@@ -504,7 +504,7 @@ def update_xpath(nsmap, xml, recprop):
                 if node1.text != recprop['value']:  # values differ, update
                     node1.text = recprop['value']
     except Exception as err:
-        print(err)
+        LOGGER.warning(err)
         raise RuntimeError('ERROR: %s' % str(err))
 
     return etree.tostring(xml)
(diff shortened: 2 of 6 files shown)


tomkralidis:
@cnavacch works for you?  I put a test for at least 4 coordinates given the point codepath doesn't use `gml:posList`.  If this works for you, I'll merge.

pycsw/ogc/gml/gml3.py
```diff
@@ -47,6 +47,11 @@ def _poslist2wkt(poslist, axisorder):
     tmp = poslist.split()
     poslist2 = []
 
+    if len(tmp) < 4:
+        msg = 'Invalid number of coordinates in geometry'
+        LOGGER.error(msg)
+        raise RuntimeError(msg)
+
     xlist = tmp[1::2]
     ylist = tmp[::2]
```
cnavacch, line 56: Will you change those two lines in another PR?

pycsw/ogc/gml/gml3.py
```diff
@@ -47,6 +47,11 @@ def _poslist2wkt(poslist, axisorder):
     tmp = poslist.split()
     poslist2 = []
 
+    if len(tmp) < 4:
```
cnavacch, line 50: Shouldn't this be `<5`, i.e. only a triangle with the starting point twice is allowed having the least number of coordinates?

cnavacch:
@tomkralidis : sorry for the late reply. All tests work fine for me.

pycsw/ogc/gml/gml3.py
tomkralidis: A triangle in WKT would look like:

```POLYGON((30 10, 40 40, 10 20, 30 10))```

File: pycsw/ogc/gml/gml3.py (hunk 1 above)
Comment (tomkralidis) [reply pycsw/ogc/gml/gml3.py:56]: The output behaviour:

```python
>>> [1, 2, 3, 4, 5, 6][1::2]
[2, 4, 6]
>>> [1, 2, 3, 4, 5, 6][::2]
[1, 3, 5]
```

pycsw/ogc/gml/gml3.py
cnavacch: Yes, that's clear, but I am still confused why the check uses 4, this would mean `48.1537 16.2825 49.1537` is not valid, but  `48.1537 16.2825 49.1537 17.2825` is valid, which is theoretically a line, not a polygon. In my opinion `30 10 40 40 10 20 30 10` is a valid polygon, but `30 10 40 40 10 20` not, so the check should be `<8` (sorry the '<5' in the comment above is wrong.)

pycsw/ogc/gml/gml3.py
cnavacch: Yes, but as already described in https://github.com/geopython/pycsw/issues/621#issue-665176298, the above expression creates inconsistencies with respect to a GML point. The following example with the point `40 10` (latitude, longitude/ X, Y) and the (invalid) polygon `40 10` (latitude, longitude/ X, Y) should illustrate this:

For a GML point, you select the coordinates as follows: 

```
xypoint = tmp.text.split()
if self.crs.axisorder == 'yx':
    self.wkt = 'POINT(%s %s)' % (xypoint[1], xypoint[0])
else:
    self.wkt = 'POINT(%s %s)' % (xypoint[0], xypoint[1]) 
```
Assuming default order ("xy") the coordinates are selected accordingly, `40` as latitude, `10` as longitude.
However, for the Polygon its different:

```
xlist = tmp[1::2]
ylist = tmp[::2]

if axisorder == 'yx':
	for i, j in zip(ylist, xlist):
		poslist2.append('%s %s' % (i, j))
else:
	for i, j in zip(xlist, ylist):
		poslist2.append('%s %s' % (i, j))
```
Assuming default order ("xy") the coordinates are selected in the wrong order, namely, `40` as longitude, `10` as latitude, because: 
```
>>>tmp_list = [40, 10]
>>>tmp_list[1::2]  # xlist
[10]
>>>tmp_list[::2] # ylist
[40]
```
but: 
```
>>>tmp_list = [40, 10]
>>>tmp_list[::2]  # xlist
[40]
>>>tmp_list[1::2] # ylist
[10]
```
would do the right job.

pycsw/ogc/gml/gml3.py
tomkralidis: The `_poslist2wkt` function works against a posList in the context of either a line or point so putting 4 would minimally satisfy the line case (2 direct positions for line, 4 direct positions for polygon).  Beyond a validating parser, there's no additional checks done (we were careful not to go to deep into GML, which would be overkill IMHO for the pycsw use case), so the expectation would be for the underlying data store to throw an exception on a bad geometry.

Having said this, I've added a fix to catch this nonetheless depending on the line or polygon geometry type.

pycsw/ogc/gml/gml3.py
tomkralidis: Fixed. I've added a test (https://github.com/tomkralidis/pycsw/blob/issue-621/tests/functionaltests/suites/default/post/GetRecords-filter-bbox-poslist.xml) which yields the same results as https://github.com/tomkralidis/pycsw/blob/issue-621/tests/functionaltests/suites/default/post/GetRecords-filter-bbox.xml.

cnavacch: Great, that's how I have expected it to work.

pycsw/ogc/gml/gml3.py
cnavacch: Okay, thank you for the clarification!

cnavacch (approved):
Thanks for all these changes!

PR #671, opened 2021-04-23
remove encoding kw from json.loads

bnavigator:
The test suite fails with python 3.9, because the `encoding` parameter [was removed](https://docs.python.org/3/library/json.html#json.loads)

AFAICT, `_test_json_result` is not called with the encoding parameter anywhere.

# Contributions and Licensing

- [x] I'd like to contribute this bugfix regarding json.loads for py3.9 to pycsw. I confirm that my contributions to pycsw will be compatible with the pycsw license guidelines at the time of contribution.


tomkralidis:
Thanks @bnavigator !

PR #680, opened 2021-05-14
make sure the repository get's loaded even after long db inactivity.

epifanio:
# Overview
Fixes a problem in establishing a connection to the db when the service has been inactive for ~120 sec. - the problem arises when the service is running in a docker swarm environment

# Related Issue / Discussion

# Additional Information

# Contributions and Licensing

### DIFF
--- a/pycsw/server.py
+++ b/pycsw/server.py
@@ -250,11 +250,7 @@ class Csw(object):
             self.requesttype = 'GET'
             self.request = wsgiref.util.request_uri(self.environ)
             try:
-                if '{' in self.request or '%7D' in self.request:
-                    LOGGER.debug('Looks like an OpenSearch URL template')
-                    query_part = self.request.split('?', 1)[-1]
-                else:
-                    query_part = splitquery(self.request)[-1]
+                query_part = splitquery(self.request)[-1]
                 self.kvp = dict(parse_qsl(query_part, keep_blank_values=True))
             except AttributeError as err:
                 LOGGER.exception('Could not parse query string')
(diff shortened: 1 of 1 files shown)


kalxas:
This re-introduces a bug that @tomkralidis has fixed in master recently

tomkralidis:
@kalxas Yes, I think in https://github.com/geopython/pycsw/pull/680/files#diff-c98be6a7c956ae4752c992b30e532b8ebc0b8de45817ab2dc284a9f555c55ca9L253-L257 .  @epifanio can you rebase/update accordingly?

epifanio:
Yes I will amend the or. Sorry it went in by mistake as it is not related to this pr.

epifanio:
@tomkralidis , I updated the PR, did I fixed the issue?

pycsw/server.py
```diff
@@ -396,8 +396,16 @@ def dispatch(self, writer=sys.stdout, write_headers=True):
             rs_cls = getattr(rs_mod, rs_clsname)
 
             try:
-                self.repository = rs_cls(self.context, repo_filter)
-                LOGGER.debug('Custom repository %s loaded (%s)', rs, self.repository.dbtype)
+                connection_done=False
+                max_attempt=0
+                while not connection_done and max_attempt<=5:
```
tomkralidis, line 401: Should we add `max_attempt` as a configurable (i.e. in `repository.max_retries`).  Then on pycsw init, we check/set in config with a default of 3.

tomkralidis (changes requested):

PR #681, opened 2021-05-14
upgrade pip in dockerfile

epifanio:
# Overview

upgrade pip before start building python dependency

# Related Issue / Discussion

# Additional Information

# Contributions and Licensing


kalxas (changes requested):
I would not introduce another RUN command, just add as && if possible

kalxas (approved):

PR #682, opened 2021-05-14
Read workers from config

epifanio:
# Overview
This PR simply add the option to read the number of workers from the  pycsw configuration for an easier deployment process
# Related Issue / Discussion

# Additional Information

# Contributions and Licensing

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -63,6 +63,8 @@ COPY --chown=pycsw \
     requirements-dev.txt \
     ./
 
+RUN pip install -U pip
+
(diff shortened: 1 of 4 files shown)


tomkralidis:
Adding @kalxas and @ricardogsilva for review.

@epifanio this PR looks like a means to specify the number of workers from config.  Can you add documentation? i.e. add the entry to `docs/configuration.rst` with explanation and default value if not specified.  We'll also need this in `default-sample.cfg` (probably commented out) as an example.

Issue #698, opened 2021-06-30
Request timeout


maurelioc:
# Description

# Environment

- operating system: Kubuntu 18.04
- Python version: 3.8
- pycsw version: 2.4.2
- source/distribution
  - [ ] git clone
  - [ ] DebianGIS/UbuntuGIS
  - [ ] PyPI
  - [ ] zip/tar.gz
  - [x] other (please specify): Docker image geopython/pycsw:2.4.2
- web server
  - [ ] Apache/mod_wsgi
  - [ ] CGI
  - [x] other (please specify): Gunicorn (docker image)

# Steps to Reproduce
Run 2 containers pycsw in separated databases.
Harvesting from one pycsw instance to the other.
Harvest request was made through python owllib csw client.
Request is stopped by disconnection from peer.

# Additional Information
The harvest works fine if the records count is low. For greater registries amounts, it interrupts the request when reaches 30 seconds of execution.

maurelioc:
According to Gunicorn docs, there is a 30 seconds timeout default for worker tasks.
It could be added to the entrypoint.py and the pycsw.cfg the possibility to config greater timeouts.

maurelioc:
I have managed to add this code to a 2.6.0 compilation and it worked just fine.

PR #699, opened 2021-07-06
Creates a timeout configuration for gunicorn

maurelioc:
# Overview
Although it is possible to specify Gunicorn's worker timeout, it is not easily passed to the application via configuration file.

# Related Issue / Discussion
#698 

# Additional Information
I have just inserted the Gunicorn's default value in pycsw.cfg and called it in entrypoint.py.

# Contributions and Licensing

- [x] I'd like to contribute gunicorn's timeout configuration to pycsw. I confirm that my contributions to pycsw will be compatible with the pycsw license guidelines at the time of contribution.


kalxas (approved):
This looks good, but we will need to immediately issue a helm chart fix for Kubernetes deployments

PR #709, opened 2021-08-20
Exception logging improvements

rcoup:
1. Bug: fix missing log argument for postgresql repository setup

2. logging: add exception info to error/warn messages.
Error tracing systems like Sentry produce much more useful events with exception contexts. While these are automatically associated for `logger.exception()` calls, that doesn't happen for other log levels. Change: Where a `.warn()`/`.error()` log call is associated with an Exception object, pass it to the logger.
Implications: in some log configurations this will log tracebacks more often

3. Use `raise ... from ...` syntax when re-raising exceptions, so cause is propagated. Again, helps error tracing.

# Contributions and Licensing

- [x] I'd like to contribute code to pycsw. I confirm that my contributions to pycsw will be compatible with the pycsw license guidelines at the time of contribution.

### DIFF
--- a/pycsw/core/admin.py
+++ b/pycsw/core/admin.py
@@ -412,7 +412,7 @@ def export_records(context, database, table, xml_dirpath):
             os.makedirs(dirpath)
         except OSError as err:
             LOGGER.exception('Could not create directory')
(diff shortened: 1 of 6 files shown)


tomkralidis:
Thanks @rcoup!

PR #742, opened 2022-04-03
Fixing CI

kalxas:
# Overview

Move to latest pyproj and revert breaking commit

# Contributions and Licensing


tomkralidis (approved):
Note that `urllib.parse.urlparse` is deprecated as of Python 3.8 (hence #735).  Merging PR to fix CI, and we can test #735 again with less moving parts.

PR #743, opened 2022-04-03
fix STAC version and collections

tomkralidis:
# Overview
Harmonizes STAC versions
# Related Issue / Discussion
None
# Additional Information
None
# Contributions and Licensing


pycsw/ogc/api/records.py
```diff
@@ -208,7 +208,7 @@ def landing_page(self, headers_, args):
         headers_['Content-Type'] = self.get_content_type(headers_, args)
 
         response = {
-            'stac_version': '1.0.0',
+            'stac_version': '1.0.0-beta.4',
```
kalxas, line 211: I think we removed the beta4 tag so that pystac_client would work/validate the response.

kalxas (approved):

kalxas (approved):

PR #746, opened 2022-04-04
safeguard Werkzeug 400 behaviour for JSON POST payloads

tomkralidis:
# Overview
Werkzeug 2.1.0 (used by Flask [in `requirements-standalone.txt]`) introduces [stricter HTTP POST JSON payload handling](https://github.com/pallets/werkzeug/issues/2339), which results in 400 errors, [specifically](https://werkzeug.palletsprojects.com/en/2.1.x/changes/#version-2-1-0):

> `Request.get_json()` will raise a 400 `BadRequest` error if the `Content-Type` header is not `application/json`. This makes a very common source of confusion more visible. [#2339](https://github.com/pallets/werkzeug/issues/2339)

This PR provides a workaround for this breaking change.

# Related Issue / Discussion
None
# Additional Information
None
# Contributions and Licensing

### DIFF
--- a/pycsw/wsgi_flask.py
+++ b/pycsw/wsgi_flask.py
@@ -161,7 +161,8 @@ def items(collection='metadata:main'):
     if 'search' in request.url_rule.rule:
         stac_item = True
 
-    return get_response(api_.items(dict(request.headers), request.json, dict(request.args),
+    return get_response(api_.items(dict(request.headers),
+                        request.get_json(silent=True), dict(request.args),
                         collection, stac_item))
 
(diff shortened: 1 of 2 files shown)


tomkralidis:
Aside: note that this would have been easier to detect/catch if the OARec and STAC tests were put forth as functional tests, not unit tests. @ricardogsilva it would be valuable to get your thoughts on putting non-CSW functional tests (as part of another discussion/issue) into the pytest suite.

kalxas (approved):

PR #800, opened 2022-10-23
skip q filter if empty

### LINKED ISSUE #799: query with q= throws invalid literal for int() with base 10: Token('INT', '')
# Description

https://demo.pycsw.org/gisdata/collections/metadata:main/items?q=

Throws

```
  File "/usr/local/lib/python3.8/site-packages/pygeofilter/parsers/ecql/parser.py", line 182, in INT
    return int(value)
ValueError: invalid literal for int() with base 10: Token('INT', '')
```
http://localhost:8000/collections/metadata:main/items and https://demo.pycsw.org/gisdata/collections/metadata:main/items?q=preserve work fine.

Expected: no filter applied, eg. all records

# Environment

docker latest (oct 2022)

pvgenuchten:
# Overview

if empty q-filter is send, don't apply a filter

# Related Issue / Discussion

fixes #799

# Additional Information

to consider to flag [empty filter](#799) error also in the cql module

# Contributions and Licensing

### DIFF
--- a/pycsw/ogc/api/records.py
+++ b/pycsw/ogc/api/records.py
@@ -569,7 +569,8 @@ class API:
                         if end != '..':
                             query_args.append(f'time_end <= "{end}"')
                 elif k == 'q':
-                    query_args.append(build_anytext('anytext', v))
+                    if v not in [None,'']:
+                        query_args.append(build_anytext('anytext', v))
                 else:
                     query_args.append(f'{k} = "{v}"')
 


kalxas (approved):

pycsw/ogc/api/records.py
```diff
@@ -569,7 +569,8 @@ def items(self, headers_, json_post_data, args, collection='metadata:main',
                         if end != '..':
                             query_args.append(f'time_end <= "{end}"')
                 elif k == 'q':
-                    query_args.append(build_anytext('anytext', v))
+                    if v is not '':
```
tomkralidis, line 572: Can we safeguard with `if v is not in [None, '']:`

tomkralidis (changes requested):
Thanks @pvgenuchten.  Minor change requested.

Issue #806, opened 2022-10-27
use pygeometa to convert iso19139 to recordgeojson?


pvgenuchten:
https://github.com/geopython/pycsw/blob/bde1825b73f50839a552708bf3fff737f4152abc/pycsw/plugins/profiles/apiso/apiso.py#L675-L697By design pycsw captures a [limited subset of iso19139](https://github.com/geopython/pycsw/blob/c380f9923dcd60691a3b24e09a0c0f21bda800a3/pycsw/plugins/profiles/apiso/apiso.py#L65-L121), this subset is then [used to populate the recordgeojson of ogcapi-records](https://github.com/geopython/pycsw/blob/c380f9923dcd60691a3b24e09a0c0f21bda800a3/pycsw/ogc/api/records.py#L941-L1040), which leads to unexpected data loss. I wonder if it would make sense pycsw prepares a recordgeojson as part of the iso19139 ingest and stores it in the database to be returned when the record is requested via ogcapi-records. The pygeometa library could be used to convert iso19139 to [recordgeojson](https://github.com/opengeospatial/ogcapi-records/blob/master/core/openapi/schemas/recordGeoJSON.yaml) via mcf.

Alternatively the pycsw datamodel could be extended to capture additional properties. For us the various contacts with their address details and quality reports with their details are most relevant. I wonder if a similar hierarchical structure could be set up as has been prepared for the [distributions/links section](https://github.com/geopython/pycsw/blob/c380f9923dcd60691a3b24e09a0c0f21bda800a3/pycsw/core/metadata.py#L1493-L1512).

PR #808, opened 2022-10-28
capture contact details on the database to be used in recordgeojson output

pvgenuchten:
# Overview

This is a suggestion on how contact details can be captured on pycsw database, so they can be used on ogc-api-records.

It maps also topiccategory and otherconstraints to geojsonrecord properties (license and theme)

# Related Issue / Discussion

#806

# Additional Information

a new field 'contacts' has been added to the pycsw database

# Contributions and Licensing

- [x] I'd like to contribute this feature to pycsw. I confirm that my contributions to pycsw will be compatible with the pycsw license guidelines at the time of contribution.


kalxas:
Thank you @pvgenuchten for the PR.
I would prefer if we avoid adding `jsons` as a new dependency.

pycsw/core/metadata.py
```diff
@@ -31,7 +31,7 @@
 #
 # =================================================================
 
-import json
+import json, jsons
```
tomkralidis, line 34: echo'ing @kalxas, prefer to use default Python json support.

pycsw/ogc/api/records.py
```diff
@@ -967,6 +967,16 @@ def record2json(record, url, collection, stac_item=False):
         'assets': {}
     }
 
+    # todo; for keywords with a scheme use the theme property
+    themes=[]
```
tomkralidis, line 971: `themes = []`

pycsw/ogc/api/records.py
```diff
@@ -967,6 +967,16 @@ def record2json(record, url, collection, stac_item=False):
...
+    themes.append({'concepts': [record.topicategory],
```
tomkralidis, line 972: Should we check if `record.topicategory` is not None first?

pycsw/ogc/api/records.py
```diff
@@ -996,6 +1006,30 @@ def record2json(record, url, collection, stac_item=False):
 
     if record.keywords:
         record_dict['properties']['keywords'] = [x for x in record.keywords.split(',')]
+    
+    if record.contacts:
+        rcnt = []
+        for cnt in json.loads(record.contacts):
+            rcnt.append({
+                'name': '{0} - {1}'.format(cnt.get('name',''),cnt.get('organization','')),
```
tomkralidis, line 1014: Format using f-strings

requirements.txt
```diff
@@ -1,4 +1,5 @@
 click
+jsons
```
tomkralidis, line 2: Remove dependency and use default Python json support.

tomkralidis (changes requested):
Great work @pvgenuchten.  See change requests.

pycsw/core/metadata.py
```diff
@@ -1381,6 +1381,7 @@ def _parse_iso(context, repos, exml):
             len(md.identification.contact) > 0):
             all_orgs = set([item.organization for item in md.identification.contact if hasattr(item, 'organization') and item.organization is not None])
             _set(context, recobj, 'pycsw:OrganizationName', ';'.join(all_orgs))
+            _set(context, recobj, 'pycsw:Contacts', jsons.dumps(md.identification.contact))
```
pvgenuchten, line 1384: introduced jsons here because json was not able to parse the contact instance from owslib, instead i can add an explicit property by property conversion (https://stackoverflow.com/questions/3768895/how-to-make-a-class-json-serializable)

requirements.txt
tomkralidis, line 2: I think we should keep to the standard library for JSON support.  cc @kalxas for comments.

tomkralidis (changes requested):

requirements.txt
kalxas: +1
we should avoid adding new dependencies if possible

kalxas (approved):

kalxas:
We need to test if the new field is populated properly when importing ISO-19115 XML

pycsw/core/config.py
```diff
@@ -175,6 +175,8 @@ def __init__(self, prefix='csw30'):
                 'pycsw:Bands': 'bands',
                 # links: list of dicts with properties: name, description, protocol, url
                 'pycsw:Links': 'links',
+                # contacts: list of dicts with properties: name, description, protocol, url
```
tomkralidis, line 178: Update comment to describe stored contacts objects

tomkralidis (changes requested):

pvgenuchten:
> We need to test if the new field is populated properly when importing ISO-19115 XML

I tested for iso, but not for others like dc

pycsw/core/config.py
kalxas: This has been resolved by latest commit

pvgenuchten:
nice, this got merged, thanx for reviewing!

PR #810, opened 2022-11-01
update pycsw-admin docs to use click attributes

pvgenuchten:
# Additional Information

https://github.com/geopython/pycsw/pull/688


pvgenuchten:
mixing up versions here, this relates to an older version, current seems fine

PR #813, opened 2022-11-05
adds externalIds and preview link  to record-geojson

pvgenuchten:
# Overview

- Record-geojson has a field externalIds [{schema:xxx,value:yyy}] to capture remote identifications of the resource. 
- Thumbnail links are included as a link of rel=preview. this PR adds both aspects to the record geojson encoding, based on existing pycsw properties (it does not introduce new pycsw field).
- While building the record json, it takes langaugecode if language is not populated

# Related Issue / Discussion

pycsw currently uses the source field for resource identification (based on md.identification.dataseturi). Future improvements: a namespace is not yet captured on pycsw database, this is for example used in wms capabilities authority:identifier (see also https://github.com/geopython/OWSLib/issues/841) but also on iso as md.identification.identifier.namespace

the preview link is build up from links with the `WWW:LINK-1.0-http--image-thumbnail` protocol, as currently used in wms-thumbnails. I currently don't populate type (image/png or jpg) maybe it can be build by parsing the url. Future work: add the thumbnail as an image to the items and item views

# Additional Information

# Contributions and Licensing

### DIFF
--- a/pycsw/core/metadata.py
+++ b/pycsw/core/metadata.py
@@ -1322,7 +1322,7 @@ def _parse_iso(context, repos, exml):
     _set(context, recobj, 'pycsw:XML', md.xml)
     _set(context, recobj, 'pycsw:MetadataType', 'application/xml')
     _set(context, recobj, 'pycsw:AnyText', util.get_anytext(exml))
-    _set(context, recobj, 'pycsw:Language', md.language)
+    _set(context, recobj, 'pycsw:Language', md.language or md.languagecode)
     _set(context, recobj, 'pycsw:Type', md.hierarchy)
     _set(context, recobj, 'pycsw:ParentIdentifier', md.parentidentifier)
     _set(context, recobj, 'pycsw:Date', md.datestamp)
@@ -1352,6 +1352,8 @@ def _parse_iso(context, repos, exml):
 
         if len(md.identification.resourcelanguage) > 0:
             _set(context, recobj, 'pycsw:ResourceLanguage', md.identification.resourcelanguage[0])
+        elif len(md.identification.resourcelanguagecode) > 0:
+            _set(context, recobj, 'pycsw:ResourceLanguage', md.identification.resourcelanguagecode[0])
 
         if hasattr(md.identification, 'bbox'):
             bbox = md.identification.bbox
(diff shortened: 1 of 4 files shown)


pycsw/core/metadata.py
```diff
@@ -1528,6 +1530,15 @@ def _parse_iso(context, repos, exml):
     except Exception as err:  # srv: identification does not exist
         LOGGER.exception('no srv:SV_ServiceIdentification links found')
 
+    if hasattr(md.identification, 'graphicoverview'):
+        for thumb in  md.identification.graphicoverview:
+            links.append({
+                'name': 'Preview',
```
tomkralidis, line 1536: Change to preview (all lower case).

tomkralidis (changes requested):

pycsw/ogc/api/records.py
```diff
@@ -1007,9 +1011,11 @@ def record2json(record, url, collection, stac_item=False):
                 'description': link['description'],
                 'type': link['protocol']
             }
-            if 'rel' in link:
+            if 'rel' in link.keys():
```
tomkralidis, line 1014: I think we don't need to specify `.keys()` explicitly?

pycsw/ogc/api/records.py
```diff
@@ -1007,9 +1011,11 @@ def record2json(record, url, collection, stac_item=False):
...
                 link2['rel'] = link['rel']
-            elif 'function' in link:
+            elif link['protocol'] == 'WWW:LINK-1.0-http--image-thumbnail':
+                link2['rel'] = 'preview'
+            elif 'function' in link.keys():
```
tomkralidis, line 1018: Same here (no explicit `.keys()` needed).

tomkralidis (changes requested):

pycsw/ogc/api/records.py
```diff
@@ -1020,9 +1023,11 @@ def record2json(record, url, collection, stac_item=False):
         'title': 'Collection',
         'name': 'collection',
         'description': 'Collection',
-        'href': f"{url}/collections/{collection}?f=json"
+        'href': f"{url}/collections/{collection}"
     })
 
+    print(vars(record))
```
tomkralidis, line 1029: remove print statement.

pycsw/ogc/api/records.py
```diff
@@ -1035,8 +1040,25 @@ def record2json(record, url, collection, stac_item=False):
                 [minx, miny]
             ]]
         }
+        record_dict['extent'] = {
```
tomkralidis, line 1043: As it happens, `extent` is no longer a part of the OARec core record model.  Can we remove?

tomkralidis (changes requested):

pycsw/ogc/api/records.py
pvgenuchten: ah i was wondering about this, so properties.extent.temporal has moved to feature.time, and properties.extent.geospatial to feature.geometry?

tomkralidis: yes.  `feature.time` and `feature.geometry` are single spatiotemporal definitions.  Any additional spatiotemporal definitions are left to downstream profiles.

tomkralidis (approved):

kalxas (approved):

Issue #814, opened 2022-11-11
how to transform a  WMS/WFS iso19139 transferoption to record.geojson?


pvgenuchten:
# Description

Following the discussion at https://github.com/opengeospatial/ogcapi-records/issues/95, to capture enough detail on a association in record.geojson so a client is able to bind to the service advertised, we need some mapping from for example the iso19139 transfer-options model to a template+variables structure as suggested in record.geojson.

iso
```xml
<gmd:MD_DigitalTransferOptions>
          <gmd:onLine>
            <gmd:CI_OnlineResource>
              <gmd:linkage>
                <gmd:URL>https://service.pdok.nl/hwh/wsdiensteninspire/wms/v1_0?service=wms&amp;request=getcapabilities</gmd:URL>
              </gmd:linkage>
              <gmd:protocol>
                <gmx:Anchor xlink:href="http://www.opengis.net/def/serviceType/ogc/wms">OGC:WMS</gmx:Anchor>
              </gmd:protocol>
              <gmd:name>
                <gco:CharacterString>kabel</gco:CharacterString>
              </gmd:name>
              <gmd:description>
                <gmx:Anchor xlink:href="http://inspire.ec.europa.eu/metadata-codelist/OnLineDescriptionCode/accessPoint">accessPoint</gmx:Anchor>
              </gmd:description>
            </gmd:CI_OnlineResource>
          </gmd:onLine>
```

record geojson
```json
{
   "rel": "map",
   "type": "image/png",
   "title": "Ozone map",
   "href": "https://geo.woudc.org/ows?service=WMS&version=1.3.0&request=GetMap&crs={crs}&bbox={bbox}&layers=totalozone&width={width}&height={height}&format=image/png",
   "templated": true,
   "variables": {
      "crs": {
         "description": "...",
         "type": "string",
         "enum": [ "EPSG:4326", "EPSG:3857" ]
      },
      "bbox": {
         "description": "...",
         "type": "array",
         "items": {
            "type": "number",
            "format": "double"
         },
         "minItems": 4,
         "maxItems": 4
      },
      "width": {
         "description": "...",
         "type": "number",
         "format": "integer",
         "minimum": 600,
         "maximum": 5000
      },
      "height": {
         "description": "...",
         "type": "number",
         "format": "integer",
         "minimum": 600,
         "maximum": 5000
      }
   }
}
```

[...]

PR #815, opened 2022-11-14
suggestion for implementing wms/wmts/wfs/... links in record.geojson

pvgenuchten:
# Overview

populate links in record.geojson for common link types using the suggested templated approach

# Related Issue / Discussion

in #814 I raised the question how record.geojson links are modelled from the pycsw datamodel for the typical wms/wfs protocols. This PR presents a suggestion for implementation

# Additional Information

A lot of conventions are adopted and discussions are taking place on how one can model a link between a iso19139 metadata record and a specific layer in ows capabilities. In general 2 approaches are common, append on the metadatalink to the service the &layers is parameter to detail the relevant layer, others adopted the &layers={onlineresource.name} convention. This PR supports both cases (but will cause problems if none of the 2 are implemented)

To identify the type of link, the [catinterop](https://github.com/OSGeo/Cat-Interop/blob/master/LinkPropertyLookupTable.csv) identifier of the protocol is assumed in link[type], else the link will be treated as a generic link

# Contributions and Licensing


tomkralidis:
@pvgenuchten can you rebase?

pycsw/ogc/api/records.py
```diff
@@ -1048,22 +1048,70 @@ def record2json(record, url, collection, stac_item=False):
...
-
-            rdl.append(link2)
+                'title': link['name'],
+                'href': link['url']
+                }
+            if link['protocol']:
+                if 'OGC:WMS' in link['protocol'].upper():
+                    link2['rel'] = 'map'
+                    link2['templated'] = 'true'
+                    # assumes link['url'] includes '&layers=...', else link['name'] contains layername(s)
```
pvgenuchten, line 1062: alternative would be to use owslib.wms130.__build_getmap_request (but challenge to replace bbox array for `{bbox}`)

pycsw/ogc/api/records.py
```diff
@@ -1048,22 +1048,70 @@ def record2json(record, url, collection, stac_item=False):
...
+                    link2['templated'] = 'true'
+                    link2['href'] = merge_qs(link['url'], {
+                        'version': '2.0.0', 'request': 'GetFeature', 'service': 'WFS', 'count': '{count}'}, {
+                        'typenames': link['name'], 'srsName': 'urn:ogc:def:crs:EPSG::4326',
+                        'outputFormat':'application/gml+xml'})
+                    link2['variables'] = { 
+                        'count': {'type': 'number', 'format': 'integer'}}
+                    link2['type'] = 'application/gml+xml'
+
+                #elif 'OSGEO:TMS' in link['protocol'].upper():
```
pvgenuchten, line 1105: or better/also to use xyz (both not in catinterop yet, see https://github.com/OSGeo/Cat-Interop/issues/42)

tomkralidis:
@pvgenuchten can you rebase/update this PR?

PR #829, opened 2022-12-09
Add option to run simpler WSGI app

epifanio:
Add option to run simpler WSGI app instead of flask  app in the docker entry-point

# Overview

Adds an entry in the configure to decide which app `wsgi` or `wsgi.flask` should run via the docker entrypoint

# Related Issue / Discussion

The flask app at the moment depends on the pycsw backend (e.g.: PostgreSQL, SQLite ) - the actual entrypoint doesn't have an option to run the simpler wsgi app to deploy a csw service

# Additional Information

# Contributions and Licensing

### DIFF
--- a/default-sample.cfg
+++ b/default-sample.cfg
@@ -45,6 +45,7 @@ gzip_compresslevel=9
 #spatial_ranking=true
 profiles=apiso
 #workers=2
+#flask_app="False"
 
 [manager]
 transactions=false
--- a/default-sample.cfg
+++ b/default-sample.cfg
@@ -45,7 +45,7 @@ gzip_compresslevel=9
 #spatial_ranking=true
 profiles=apiso
 #workers=2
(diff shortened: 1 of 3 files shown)


kalxas (changes requested):
Thanks for the PR, please review the comments.

epifanio:
@tomkralidis - can we look into this?

frafra:
I tried to add some comments, since feedback has been requested on the gitter channel.

I am trying to get my head around the Dockerfile. Why the code to start pycsw is in an entrypoint in the first place?

epifanio:
In this change i simply bypass the db settings/connection as it is not required to run the CSW services. Which makes the code more likely a standard docker entry point. Perhaps the db settings should go into the flask app itself?

tomkralidis:
Closing in lieu of #907

PR #872, opened 2023-06-14
Update STAC Common Metadata datetime and keywords fields

kalxas:
# Overview

Updating pycsw STAC parser with STAC Common Metadata fields datetime and keywords
https://github.com/radiantearth/stac-spec/issues/1187
https://github.com/radiantearth/stac-spec/pull/1227
https://github.com/radiantearth/stac-spec/blob/master/item-spec/common-metadata.md#date-and-time

# Contributions and Licensing

### DIFF
--- a/pycsw/core/metadata.py
+++ b/pycsw/core/metadata.py
@@ -1887,7 +1887,7 @@ def _parse_stac_resource(context, repos, record):
         if 'collection' in record:
             _set(context, recobj, 'pycsw:ParentIdentifier', record['collection'])
 
-    if stac_type == 'Collection':
+    if stac_type == 'Collection' or stac_type == 'Catalog':
         if 'keywords' in record:
             keywords = record['keywords']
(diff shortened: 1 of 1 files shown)


pycsw/core/metadata.py
```diff
@@ -1887,11 +1887,23 @@ def _parse_stac_resource(context, repos, record):
         if 'collection' in record:
             _set(context, recobj, 'pycsw:ParentIdentifier', record['collection'])
 
-    if stac_type == 'Collection':
+    if stac_type == 'Collection' or stac_type == 'Catalog':
```
tomkralidis, line 1890:
```python
    if stac_type in ['Collection', 'Catalog']:
```

tomkralidis (changes requested):

pycsw/core/metadata.py
kalxas: Fixed, thanks

Issue #850, opened 2023-03-14
linkage between records


pvgenuchten:
# Description

Does pycsw offer a capability to establish links between records? These links would then be advertised in the link section of an ogc-api-records representation. Common linkages in metadata are the parent-child relation and the service-dataset link. However iso19139 describes a number of alternative linkages for example via the  [aggregationInfo](http://www.datypic.com/sc/niem21/e-gmd_aggregationInfo-1.html) element.

pvgenuchten:
initial research, db model has `parentidentifier` and `relation`, relation is populated from [iso.aggregationinfo](https://github.com/geopython/pycsw/blob/3243c43c613d1fc951f3aacd6c00cbbeba119a6a/pycsw/core/metadata.py#L1406) as [simpleText](https://github.com/geopython/OWSLib/blob/bc88ee1a6509746cff1cccfbad6e6501bae4c256/owslib/iso.py#L406)

Both parentidentifier and relation contain uri's/uuid's to linked records. OGCAPI-records expects URI's. 

Suggestion would be to use item-url as link to record on records links object, unless identifier is already a url (startswith(http))

pvgenuchten:
#880 optimizes the orgapi-records representation of a record, it does not cover other aspects in this issue such as parentidentifier, aggregationinfo etc, suggest to keep issue open, or shall i create targeted issues?

tomkralidis:
Thanks @pvgenuchten; target issues would be great

PR #880, opened 2023-08-04
add record links to ogc-api records item

pvgenuchten:
# Overview

this introduces values from rec.relation and rec.parentidentifier as links on an ogcapi-records item, it is a first iteration of the changes suggested in #850

q: can parentidentifier/relation have multiple links (separated by ',')? in that case i need to consider that case in code
q: should I add link with ?f=html as well as ?f=json
q: makes sense to query db for existince of uuid, to include also title/abstract/lang as part of this PR?

# Related Issue / Discussion

#850

# Contributions and Licensing

### DIFF
--- a/pycsw/ogc/api/records.py
+++ b/pycsw/ogc/api/records.py
@@ -33,7 +33,7 @@ from configparser import ConfigParser
 import json
 import logging
 import os
-from urllib.parse import urlencode
+from urllib.parse import urlencode, quote
 
 from pygeofilter.parsers.ecql import parse as parse_ecql
 from pygeofilter.parsers.cql2_json import parse as parse_cql2_json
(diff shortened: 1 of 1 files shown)


tomkralidis:
Thanks @pvgenuchten.  Comments on your questions below:

> q: can parentidentifier/relation have multiple links (separated by ',')? in that case i need to consider that case in code

No, parentidentifier/relation is a single record

> q: should I add link with ?f=html as well as ?f=json

We can go with JSON for now.

> q: makes sense to query db for existince of uuid, to include also title/abstract/lang as part of this PR?

Probably overkill, so fine for now.

pvgenuchten:
nice! stac browser now shows the [ogcapi-records content](https://radiantearth.github.io/stac-browser/#/external/demo.pycsw.org/cite/collections/metadata:main)

image

Issue #879, opened 2023-08-03
record not found on full title, if title contains 'of' or 'in'


pvgenuchten:
# Description
If the title of a record contains the word 'of', 'for' or 'in' (and probably all stopwords), the record will not be found if you search on full title using q search in ogcapi-records

I found this behaviour at this record (having postgres backend):
https://rwanda.lsc-hubs.org/collections/metadata:main/items?q=Mycorrhizal+colonization+major+banana+genotypes+six+East+African+environments vs
https://rwanda.lsc-hubs.org/collections/metadata:main/items?q=Mycorrhizal+colonization+of+major+banana+genotypes+in+six+East+African+environments

Does it make sense to strip stopwords from q before querying the database?

pvgenuchten:
actually when testing with postgres, i got this error message: 'column anytext_tsvector does not exist', so apparently the mechanism of full text search seems implemented, but somehow not activated, after creating the columns and trigger, full text search on title works without problems, so i guess i should rename this issue to: admin.py does not create tsvector tables at db-setup...?

pvgenuchten:
no creation of ts-vector columns/triggers seem related to:
- no postgis
- no plpythonu
when deploying pycsw on a postgis enabled db, all seem fine, ts-vector columns created and populated

pvgenuchten:
Problem is that for each keyword in q the full text search is triggered, however a fts on a stop word fails

plainto_tsquery('english', '%Aerial%') @@ anytext_tsvector
plainto_tsquery('english', '%the%') @@ anytext_tsvector
plainto_tsquery('english', '%Photos%') @@ anytext_tsvector

in stead should be
plainto_tsquery('english', 'Aerial & the & Photos') @@ anytext_tsvector

tomkralidis:
Some further analysis on FTS.  Consider a record with `anytext` field containing the terms `auxiliary` and `https://stac-extensions.github.io/datacube/v2.0.0/schema.json`:

```
pycsw=# select identifier from records where anytext like '%auxiliary%' limit 1;
                  identifier                   
-----------------------------------------------
 ESACCI-LC-L4-LCCS-Map-300m-P1Y-2009-v2.0.7cds
(1 row)

pycsw=# select identifier from records WHERE to_tsquery('english', 'auxiliary') @@ anytext_tsvector;
                  identifier                   
-----------------------------------------------
 ESACCI-LC-L4-LCCS-Map-300m-P1Y-2009-v2.0.7cds
(1 row)

[...]

PR #881, opened 2023-08-04
remove stopwords from anytext filter

pvgenuchten:
# Overview

initial rough implementation to improve #879 

q: seems only relevant to postgres, because fulltextsearch does not exist in plain sqlite.
q: it's probably better to retrieve stopwords from postgres, to prevent asynchronity, but also support stopwords for other languages ( i have not found how)

# Related Issue / Discussion

#879

# Additional Information

# Contributions and Licensing

### DIFF
--- a/pycsw/ogc/api/records.py
+++ b/pycsw/ogc/api/records.py
@@ -1164,6 +1164,8 @@ def build_anytext(name, value):
 
     predicates = []
     tokens = value.split()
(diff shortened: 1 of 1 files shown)


pvgenuchten:
actually when testing with postgres, i got this error message: 'column anytext_tsvector does not exist', so apparently the mechanism of full text search seems implemented, but somehow not activated, after creating the columns and trigger, full text search on title works without problems, so i guess i should rename this issue to: admin.py does not create tsvector tables at db-setup...?

PR #882, opened 2023-08-04
[WIP] use a single fts query, not by word

pvgenuchten:
# Overview

on anytext searches with multiple words, instead of creating multiple where clauses this creates a single where clause with a ts-query having fuzzy matching for each term. this is not valid CQL, but final results are good (on postgres, for sqlite an alternative solution is needed). I would like to hear from you what you think, are there other ways to bypass the cql syntax, because ts-query doesn't really fit with CQL...

# Related Issue / Discussion

#879

# Additional Information

# Contributions and Licensing

### DIFF
--- a/pycsw/core/pygeofilter_evaluate.py
+++ b/pycsw/core/pygeofilter_evaluate.py
@@ -66,7 +66,7 @@ class PycswFilterEvaluator(SQLAlchemyFilterEvaluator):
         if (str(lhs.prop) == 'dataset.anytext' and
                 self._pycsw_dbtype.startswith('postgres')):
             LOGGER.debug('Kicking into PostgreSQL FTS mode')
-            return text(f"plainto_tsquery('english', '{node.pattern}') @@ anytext_tsvector")  # noqa
+            return text(f"to_tsquery('english', '{node.pattern}') @@ anytext_tsvector")  # noqa
         else:
             LOGGER.debug('Default ILIKE behaviour')
             return filters.like(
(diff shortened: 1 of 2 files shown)


pvgenuchten:
test fails on sqlite, sqlite multiple word search is now broken         

```
params = {'q': 'Lorem dolor'}
        content = json.loads(api.items({}, None, params)[2])
>       assert content['numberMatched'] == 1
```

pycsw/core/pygeofilter_evaluate.py
```diff
@@ -66,7 +66,7 @@ def ilike(self, node, lhs):
         if (str(lhs.prop) == 'dataset.anytext' and
                 self._pycsw_dbtype.startswith('postgres')):
             LOGGER.debug('Kicking into PostgreSQL FTS mode')
-            return text(f"plainto_tsquery('english', '{node.pattern}') @@ anytext_tsvector")  # noqa
+            return text(f"to_tsquery('english', '{node.pattern}') @@ anytext_tsvector")  # noqa
```
tomkralidis, line 69: Can we keep `plainto_tsquery`?  This is what is used in CSW 2/3 support, so it would be valuable for consistency.

Having said this, as we plan to update pycsw to exclusively use pygeoafilter for all queries in a future release, that would be a good opportunity to update PostgreSQL FTS to use `to_tsquery` accordingly.

tomkralidis (changes requested):

pycsw/core/pygeofilter_evaluate.py
pvgenuchten: Plainto query does not support fuzzy matching, also consider that #881 is also still an alternative, I closed it because it is less optimal (from multilingual perspective)

pvgenuchten:
closing for now, to be discussed soon

oapi is moving to non full-text-search

introduce full-text-search as a profile?

PR #901, opened 2023-09-21
Filter anytext ilike

tomkralidis:
# Overview
- When OGC API - Records receives a CQL request, detect wildcards and, if found, suppress FTS queries
- update OpenAPI server info to pycsw version (which is different than the OpenAPI version per se)
# Related Issue / Discussion
None
# Additional Information
None
# Contributions and Licensing


pvgenuchten:
Hi Tom, check also #881 and #882, hopefully we can solve both cases at once

PR #907, opened 2023-09-29
run PostgreSQL/SQLite workflow only if configured

tomkralidis:
# Overview
Replaces #829.  Only execute pycsw RDBMS codepath is pycsw is configured with PostgreSQL or SQLite.

cc @epifanio for validation (and whose pycsw backend is SOLR).

# Related Issue / Discussion
None
# Additional Information
None
# Contributions and Licensing

### DIFF
--- a/docker/entrypoint.py
+++ b/docker/entrypoint.py
@@ -54,7 +54,7 @@ from sqlalchemy.exc import ProgrammingError
 from pycsw.core import admin
 from pycsw.core.util import EnvInterpolation
 
-logger = logging.getLogger(__name__)
+LOGGER = logging.getLogger(__name__)
 
 
 def launch_pycsw(pycsw_config, workers=2, reload=False):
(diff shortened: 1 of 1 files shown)


ricardogsilva (approved):
I guess a simpler

```python
if db_handler is not None:
    db_handler(...)
```

would work here instead of the rest of the changes, but regardless I'm definitely +1!

Just a small question: since this change is being made in order to accommodate SOLR backends, don't these also require some sort of handling too? Even if just waiting to see if the SOLR service is responding?

In any case, its +1 from me

tomkralidis:
> I guess a simpler
> 
> ```python
> if db_handler is not None:
>     db_handler(...)
> ```
> 
> would work here instead of the rest of the changes, but regardless I'm definitely +1!
> 
> Just a small question: since this change is being made in order to accommodate SOLR backends, don't these also require some sort of handling too? Even if just waiting to see if the SOLR service is responding?
> 
> In any case, its +1 from me

+1, I've made the change above.  Thanks @ricardogsilva !

Thinking more, perhaps `pycsw.admin.setup_db` should turn into `pycsw.core.repository.setup`, with the default being RDBMS, to which other repository plugins can update accordingly.  Then `docker/entrypoint.py` simply calls a repository function.  Thoughts?

Having said this, my next task further abstracting the OGC API - Records code away from RDBMS, at which point the above will likely be addressed.

@epifanio: once you also can validate this PR, I'll merge accordingly.

kalxas (approved):

epifanio:
This PR handles the DB settings in the entrypoint but the problem in "not using a DB as backend" persist when running the CSW Flask APP.  As the FLASK app still depends on a DB connection. 

In the Flask APP, the  `API` class from `pycsw.ogc.api.records`: 
[`api_ = API(APP.config['PYCSW_CONFIG'])`](https://github.com/geopython/pycsw/blob/e1abe9bc9de68f5c25061be031269752a79d384e/pycsw/wsgi_flask.py#L55C12-L55C12)  will make a new call to the DB from which it will build the   [`query_mappings`](https://github.com/geopython/pycsw/blob/e1abe9bc9de68f5c25061be031269752a79d384e/pycsw/ogc/api/records.py#L145)

epifanio:
I tested with the [last commit](https://github.com/geopython/pycsw/pull/907/commits/a1efd39f283ba9292a8bb72269381332e9f36de2) it now works as expected, by running the WSGI app instead of the default Flask app.

@tomkralidis thanks!

Issue #782, opened 2022-09-21
Bug: CSW generates wrong DB queries while nesting logical operators (AND, OR)


asafMasa:
# Description
While using a nested query we experienced an issue which we think is a bug with the generated query in the CSW.

The Query we used:
`
<csw:Constraint version="1.1.0">
      <Filter xmlns="http://www.opengis.net/ogc">
      	<Or>
      		<And>
		        <PropertyIsLike wildCard="%" singleChar="_" escapeChar="\\">
		          <PropertyName>id</PropertyName>
		          <Literal>e6c331ca-f077-452c-9811-25fc9d99acd2</Literal>
		        </PropertyIsLike>
		        <PropertyIsLike wildCard="%" singleChar="_" escapeChar="\\">
		          <PropertyName>version</PropertyName>
		          <Literal>1</Literal>
		        </PropertyIsLike>
	        </And>
	        <And>
		        <PropertyIsLike wildCard="%" singleChar="_" escapeChar="\\">
		          <PropertyName>externalId</PropertyName>
		          <Literal>test</Literal>
		        </PropertyIsLike>
		        <PropertyIsLike wildCard="%" singleChar="_" escapeChar="\\">
		          <PropertyName>externalId</PropertyName>
		          <Literal>test</Literal>
		        </PropertyIsLike>
	        </And>
        </Or>
      </Filter>
    </csw:Constraint>
`

Pycsw Log:
[...]

PR #932, opened 2023-12-27
fix parse of nested multiple binary ops (#782)

barelsh:
# Overview
Bug fix - binary operations parsed incorrectly when used multiple times in a nested constraint

# Related Issue / Discussion
 Issue #782 
# Additional Information
The inner queries list was not clened between iterations, so the expressions were appended on the previous iterations' expressions, resulting in a wrong SQL query built. (first properties were duplicated into the second condition)

# Contributions and Licensing

- [V] I'd like to contribute bugfix #782 to pycsw. I confirm that my contributions to pycsw will be compatible with the pycsw license guidelines at the time of contribution.


kalxas:
Thank you @barelsh for your contribution.
We will review in the following days

PR #941, opened 2024-01-07
add Dockerfile CVE check, update vulnerability scanning to use trivy action

tomkralidis:
# Overview
add Dockerfile CVE check, update vulnerability scanning to use trivy action
# Related Issue / Discussion
None
# Additional Information
None
# Contributions and Licensing

### DIFF
--- a/.github/workflows/vulnerabilities.yml
+++ b/.github/workflows/vulnerabilities.yml
@@ -14,27 +14,33 @@ on:
       - released
 
 jobs:
-  clone:
-    runs-on: ubuntu-22.04
-    steps:
-      - name: Setup Python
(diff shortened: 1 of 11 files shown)


kalxas:
Testing only the docker image for vulnerabilities could give the impression that all deployment methods are tested as well.

In order to pass the tests here, we would also need to change the base docker image and put effort to catch up with OS security patches.

tomkralidis:
PR updated.  Note that upgrading `Dockerfile` to `python:3.10-slim-buster` and adding `apt upgrade` passes the trivy image scan.

kalxas (approved):

Issue #948, opened 2024-01-30
at items html do not include json for map display, but fetch it with ajax


pvgenuchten:
# Description

Especially if page size is large (50+) a large geojson is [included in html](https://github.com/geopython/pycsw/blob/ac81aa8f3c362b8fa76439bb94189c943cffb267/pycsw/ogc/api/templates/items.html#L148), this decreases page load, and potentially causes problems, instead i suggest to fetch the records separately with an ajax request.

PR #949, opened 2024-01-30
fetches records for map display as ajax

pvgenuchten:
# Overview

The PR suggests to fetch records as ajax for loading on map, instead of rendering geojson inline

# Related Issue / Discussion

#948

# Contributions and Licensing

### DIFF
--- a/pycsw/ogc/api/templates/items.html
+++ b/pycsw/ogc/api/templates/items.html
@@ -137,29 +137,22 @@
 
 {% block extrafoot %}
 {% if data['features'] %}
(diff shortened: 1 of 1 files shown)


tomkralidis:
Thanks @pvgenuchten.  I guess the question here is weighing 2 HTTP requests vs 1 (with a bigger response length because of the data inline in JavaScript).  Adding @kalxas and @ricardogsilva for any comments.

pvgenuchten:
closing this for limited interest

Issue #947, opened 2024-01-29
federated catalogues None error


pvgenuchten:
# Description

if I set federated catalogues to None, i run into this error on /collections/metadata:main, what is the best way to deactivate federated catalogues?

![image](https://github.com/geopython/pycsw/assets/299829/839f4c43-8957-4235-8fa0-e0e1ee8302dc)

# Environment

- operating system: docker/ubuntu
- Python version:
- pycsw version: latest

pvgenuchten:
seems better to remove the full key in stead of making it empty

https://github.com/geopython/pycsw/blob/ac81aa8f3c362b8fa76439bb94189c943cffb267/pycsw/ogc/api/records.py#L1021

tomkralidis:
Thanks @pvgenuchten. Would the following work instead?

```diff
diff --git a/pycsw/ogc/api/records.py b/pycsw/ogc/api/records.py
index 83c3179e..59c077ac 100644
--- a/pycsw/ogc/api/records.py
+++ b/pycsw/ogc/api/records.py
@@ -1018,7 +1018,7 @@ class API:
         }
 
         if collection_name == 'metadata:main':
-            if 'federatedcatalogues' in self.config:
+            if self.config.get('federatedcatalogues') is not None:
                 LOGGER.debug('Adding federated catalogues')
                 collection_info['federatedCatalogues'] = []
                 for fc in self.config.get('federatedcatalogues', []):
```

pvgenuchten:
yes it would work, updating pr

PR #950, opened 2024-01-30
do not fail if fedcats is empty in config

pvgenuchten:
# Overview

in python, if foo['bar'] is None:
```
foo.get('bar',[]) == None
```
therefore add a None check, before calling it, probable there is a more efficient way to do it...

# Related Issue / Discussion

#947 

# Contributions and Licensing


pvgenuchten:
updated as suggested @tomkralidis

PR #959, opened 2024-02-26
proposal to move title/abstract from item-table to page header

pvgenuchten:
# Overview

this moves the type, title and abstract properties from the table to the header of the html representation for readability

image

It also uses the item title on html-head-title 

# Contributions and Licensing

### DIFF
--- a/pycsw/ogc/api/templates/item.html
+++ b/pycsw/ogc/api/templates/item.html
@@ -1,5 +1,5 @@
 {% extends "_base.html" %}
-{% block title %}{{ super() }} {{ data['id'] }} {% endblock %}
+{% block title %}{{ data.get('properties',{}).get('title',data['id']) }} - {{ super() }}{% endblock %}
 {% block extrahead %}
     <link rel="stylesheet" href="https://unpkg.com/leaflet@1.7.1/dist/leaflet.css"/>
     <script src="https://code.jquery.com/jquery-3.6.0.js"></script>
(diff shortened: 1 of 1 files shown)


tomkralidis (approved):

pycsw/ogc/api/templates/item.html
```diff
@@ -41,76 +46,77 @@ <h2>{{ data['id'] }}</h2>
...
-                  {% endfor %}
-                  </ul>
-                {% endfor %}
-                </td>
-              {% elif key == 'externalIds' %}
-                <td>
-                  <ul>
-                    {% for id in value %}
-                    <li>{{ id['scheme'] | urlize }} {{ id['value'] | urlize }}</li>
+              {% if key not in ['title', 'description', 'type'] %}
```
kalxas, line 49: I think we should still keep those fields included in the properties, even if they are rendered above

kalxas (changes requested):

pycsw/ogc/api/templates/item.html
pvgenuchten: isn't that a duplication? mostly for description, that field can have a lot of text

(ok for type and title, maybe first 100chars of description?)

kalxas: The user needs to realize that those fields are included in the JSON properties. Yes, this leads to duplication. Thanks for the update!

kalxas (approved):

pvgenuchten:
Thanx for review!

PR #960, opened 2024-02-26
Allow slash in id

### LINKED ISSUE #848: ogc api records 404 if uuid contains '/' or '%2F'
# Description

Some communities tend to place a doi in metadata identifier e.g. `10.5281/zenodo.4088113`
If you navigate to this item using /collections/metadata:main/items/10.5281/zenodo.4088113, then a 404 is returned, same error occurs with /collections/metadata:main/items/10.5281%2Fzenodo.4088113 (urlencoded).

I kind of expected option 2 to work fine, in which case we would have to make sure we always encode the '/' to %2F, alternatives could be to prevent '/' in identifiers by substituting to '-' or throw an error on insert.

pvgenuchten:
# Overview

if id contains a /, a 404 is rendered, by introducing `path` keyword in blueprint, this is solved (copied from pygeoapi)

image

Resolves #848

# Contributions and Licensing

### DIFF
--- a/pycsw/ogc/api/templates/item.html
+++ b/pycsw/ogc/api/templates/item.html
@@ -1,5 +1,5 @@
 {% extends "_base.html" %}
-{% block title %}{{ super() }} {{ data['id'] }} {% endblock %}
+{% block title %}{{ data.get('properties',{}).get('title',data['id']) }} - {{ super() }}{% endblock %}
 {% block extrahead %}
     <link rel="stylesheet" href="https://unpkg.com/leaflet@1.7.1/dist/leaflet.css"/>
     <script src="https://code.jquery.com/jquery-3.6.0.js"></script>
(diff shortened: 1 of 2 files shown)


pvgenuchten:
Apperently this did not fully solve the case, no plain 404 is rendered anymore, but for some cases a  not found exception is displayed

image

see: https://demo.pycsw.org/gisdata/collections/metadata:main/items/http://capita.wustl.edu/DataspaceMetadata_ISO/CIRA.VIEWS.COf.xml

tomkralidis:
Weird, this works for me both locally as well as running demo.pycsw.org locally via docker-compose.  @kalxas: can we validate the latest image is being pulled on demo.pycsw.org?

pvgenuchten:
indeed, i think that is the problem, version has not been updated yet

kalxas:
```
/opt/demo.pycsw.org$ sudo docker pull geopython/pycsw:latest
latest: Pulling from geopython/pycsw
Digest: sha256:47ef7fab7de761e0798555eb5f872cce75696b0f7fcf474e5a73137d0f17ae4c
Status: Image is up to date for geopython/pycsw:latest
docker.io/geopython/pycsw:latest
```

Issue #980, opened 2024-06-20
STAC API failing pystac_client validation when STAC Item is generated from ISO 19115 XML


kalxas:
# Description

When a metadata record is ingested from ISO 19115 the time information is stored in the model using `time_begin` and `time_end` columns.

When this record is pulled from the STAC API, a record is generated first:
https://github.com/geopython/pycsw/blob/master/pycsw/ogc/api/records.py#L1289-L1296

Then it is modified into a STAC Item here:
https://github.com/geopython/pycsw/blob/master/pycsw/stac/api.py#L332-L387
and here:
https://github.com/geopython/pycsw/blob/master/pycsw/stac/api.py#L389-L412

What happens is that the final STAC Item is left with a `time` property and not `datetime` or `start_datetime`,`end_datetime` as required by STAC specification

e.g. https://resource-catalogue.develop.eoepca.org/stac/collections/metadata:main/items/
https://resource-catalogue.develop.eoepca.org/stac/collections/metadata:main/items/S2MSI2A

kalxas:
Fixed

PR #981, opened 2024-06-20
fix items STAC datetime (#980)

tomkralidis:
# Overview
Adds STAC `datetime` or `start_datetime`/`end_datetime` to items responses in STAC mode.
# Related Issue / Discussion
#980
# Additional Information
None
# Contributions and Licensing

### DIFF
--- a/pycsw/ogc/api/records.py
+++ b/pycsw/ogc/api/records.py
@@ -1295,6 +1295,13 @@ def record2json(record, url, collection, mode='ogcapi-records'):
         else:
             record_dict['time'] = record.time_begin
 
+    if mode == 'stac-api':
+        if record.time_begin is not None or record.time_end is not None:
+            record_dict['properties']['start_datetime'] = record.begin_time
+            record_dict['properties']['end_datetime'] = record.end_time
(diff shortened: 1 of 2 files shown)


pycsw/ogc/api/records.py
```diff
@@ -1295,6 +1295,13 @@ def record2json(record, url, collection, mode='ogcapi-records'):
         else:
             record_dict['time'] = record.time_begin
 
+    if mode == 'stac-api':
+        if record.time_begin is not None or record.time_end is not None:
```
kalxas, line 1299: I think we need `and` here, the spec says:
`So if you use start_datetime you need to add end_datetime and vice-versa. Both fields are also REQUIRED if the datetime field is set to null. The datetime property in a STAC Item and these fields are not mutually exclusive.`
If only one exists, we just set `datetime` from it.

kalxas: https://github.com/radiantearth/stac-spec/blob/master/item-spec/common-metadata.md#date-and-time-range

kalxas (approved):

PR #987, opened 2024-08-03
[WIP] fail when repository cannot init

tomkralidis:
# Overview
This PR causes pycsw to fail on repository initialization if a connection cannot be made to the underlying metadata repository (current behaviour fails silently on repository init, and provides misleading error on subsequent interactions with `self.repository`).

# Related Issue / Discussion
None
# Additional Information
None
# Contributions and Licensing

### DIFF
--- a/pycsw/server.py
+++ b/pycsw/server.py
@@ -425,8 +425,8 @@ class Csw(object):
                     except:
                         LOGGER.debug(f'Repository not loaded retry connection {max_attempts}')
                         max_attempts += 1
-                if not connection_done:
-                    raise
+                        if max_attempts == self.max_retries:
+                            raise
             except Exception as err:
                 msg = 'Could not load repository (local): %s' % err
                 LOGGER.exception(msg)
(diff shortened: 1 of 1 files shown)


tomkralidis:
@epifanio see the 3 commits in this PR which represent different approaches that you can test in your deployment when you have a chance:
- 2391ff9cf79eaf58a6eb42a43cb8f64a3d447c96: IMHO this will not work because the repo init will fail on the first load
- 0bd11f5db2f2f7aa6ec00f0b3dd72eaef6ad4aa5: pycsw fails fast, and the server will not start (hard stop).  Not sure if this is too strict of an approach
- a2899098531aa02171018dd2f3a4abbc29a7bb07: raises on last attempt, throws error (server will not start, hard stop)

cc @kalxas @ricardogsilva

DK96-OS:
Only 0bd11f5 makes sense in the while loop context. The other commits would be better if they removed the while loop.

tomkralidis:
Still valid/in scope.

tomkralidis:
still in scope

PR #1014, opened 2024-09-12
adds the contactinfo from dedicated role fields

pvgenuchten:
adds the contactinfo from dedicated role fields, if the role does not…already exists in contacts

# Overview

the ogcapi- records json fully depends on record.contacts, however in some cases records.contacts is not populated (for example from dublin core), then fall back to the explicit role fields 'creator', 'publisher', 'contributor'. 

### DIFF
--- a/pycsw/core/metadata.py
+++ b/pycsw/core/metadata.py
@@ -1651,6 +1651,8 @@ def _parse_dc(context, repos, exml):
 
     md = CswRecord(exml)
 
+    print(md)
+
     if md.bbox is None:
         bbox = None
     else:
--- a/pycsw/core/metadata.py
+++ b/pycsw/core/metadata.py
@@ -1651,8 +1651,6 @@ def _parse_dc(context, repos, exml):
 
     md = CswRecord(exml)
 
-    print(md)
-
     if md.bbox is None:
         bbox = None
     else:
--- a/pycsw/ogc/api/records.py
+++ b/pycsw/ogc/api/records.py
@@ -1196,9 +1196,11 @@ def record2json(record, url, collection, mode='ogcapi-records'):
 
     if record.contacts not in [None, '', 'null']:
         rcnt = []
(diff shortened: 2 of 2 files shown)


pycsw/ogc/api/records.py
```diff
@@ -1215,6 +1217,13 @@ def record2json(record, url, collection, mode='ogcapi-records'):
                     })
                 except Exception as err:
                     LOGGER.exception(f"failed to parse contact of {record.identifier}: {err}")
+            for r2 in "creator,publisher,contributor".split(","): # match role-fields with contacts
```
tomkralidis, line 1220: I think `record` is an object here (not a dict)?  Perhaps the below can work?

```python
            for r2 in ['creator', 'publisher', 'contributor']:  # match role-fields with contacts
                if r2 not in roles and getattr(record, r2, None) not in [None, '']:
                    rcnt.append({
                        'organization': getattr(record, r2),
                        'roles': [r2]
                    })
```

tomkralidis (changes requested):

pycsw/ogc/api/records.py
pvgenuchten: thanx tom, you're right, now fixed and tested

pycsw/core/metadata.py
```diff
@@ -1651,6 +1651,8 @@ def _parse_dc(context, repos, exml):
 
     md = CswRecord(exml)
 
+    print(md)
```
tomkralidis, line 1654: Remove print statement.

tomkralidis (changes requested):

pycsw/core/metadata.py
pvgenuchten: oops

pvgenuchten:
this code is not executed if contacts is None, which was actually the goal to introduce it

PR #1029, opened 2024-10-03
add uri reference which is populated from dataseturi or uricode

pvgenuchten:
# Overview

This adds datasetUri / uricode as a uri property to pycsw, then adds it as a link to ogc-api records

# Related Issue / Discussion

https://github.com/geopython/OWSLib/issues/946

# Additional Information

# Contributions and Licensing


pvgenuchten:
welcoming your thoughts, i still have to fix tests (demo server is failing now and then)

PR #1067, opened 2025-01-09
split keywords by , in facet, then recalculate

### LINKED ISSUE #1065: [ogcapi-records] keywords facet should split keywords to list
# Description

at the moment the keywords facet does not manage the case if the column contains multiple keywords
the combination of keywords is returned with a count, expected is that the list is split to individual keywords and keywords are counted again on occurence.

now:
```
tree: 5
leaf: 6
tree,leaf: 2
```

expected
```
tree: 7
leaf: 8
```

Screenshot:
image

pvgenuchten:
# Overview

This PR introduces a procedure which splits keywords by ',' and recalculates totals after db query
Similar mechanism could be used for other fields which are ,-separated
I see 2 alternatives for this PR:
- keywords not to be stored as concatenated string, but actual list
- introduce an in-db procedure to split the keywords-string as part of the facet query

# Related Issue / Discussion

resolves #1065 

# Additional Information

image

# Contributions and Licensing

### DIFF
--- a/pycsw/ogc/api/records.py
+++ b/pycsw/ogc/api/records.py
@@ -1098,6 +1098,22 @@ class API:
                     'value': fq[0],
                     'count': fq[1]
                 })
+            
+            if facet in ['keywords']:
+                splitkws = {}
+                for k in facets_results[facet]['buckets']:
+                    if 'value' in k.keys() and k['value'] not in [None,''] and int(k['count']) > 0:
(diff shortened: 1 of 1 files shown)


pvgenuchten:
we found an interesting side effect here (potentially related)

if a term contains a space, the aggregated statistics provide a much higher number then when you filter by that term

image
after filtering by 'water holding capacity'
image

Issue #974, opened 2024-05-13
Prev-next buttons on item page


pvgenuchten:
# Description

I’d like to see, and I hear it from users, an option to navigate to the next item of the search result from the item view of a record.

probably easiest would be to store a list of UUID’s of the search result page in a cookie and navigate through that list until the end.

mkvaishnavi:
Hi, I’d like to work on this issue. I plan to start with the cookie-based UUID list approach mentioned above.

PR #1187, opened 2025-12-27
Add prev/next buttons navigation support for individual items

mkvaishnavi:
Related to #974 
While browsing individual items, it’s currently not easy to move to the previous or next record without going back to the items list.

This PR improves the HTML item view by adding Previous and Next navigation buttons below the item details table. The buttons reuse the existing pagination links already included in the response (rel="prev" / rel="next"), so no new pagination logic is introduced.

Added Prev / Next buttons to the item HTML page
Uses existing pagination metadata
Does not affect JSON output or API behavior

This makes browsing records more intuitive while keeping the backend behavior unchanged.

I confirm that my contributions to pycsw are compatible with the pycsw license guidelines.

### DIFF
--- a/pycsw/ogc/api/records.py
+++ b/pycsw/ogc/api/records.py
@@ -948,6 +948,37 @@ class API:
             record = self.repository.query_ids([item])[0]
             response = record2json(record, self.config['server']['url'],
                                    collection, self.mode)
+            query = self.repository.session.query(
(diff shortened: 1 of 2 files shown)


tomkralidis:
Thanks for this PR.  My main concern is introducing an additional query to the underlying repository.  Pining @geopython/pycsw for thoughts.

